Issue #92, opened 2016-10-02
Qn: Breadcrumb Trail


bsr203:
I read the section http://grahammendick.github.io/navigation/documentation/breadcrumb-trail.html

as I understood, this provides more of a navigation back functionality through selected states within history. Not sure this is same as [breadcrumb navigation](https://en.wikipedia.org/wiki/Breadcrumb_(navigation)). Say user is provided a direct link to `/author/xyz/book/123`, and if I have Breadcrumb trail ( static link) to `/author/{id}`, does it populate correctly? I don't see the relevance of distance there, unless a way to specify hierarchy.

I could build my trail with `NavigationLink`, but want to make sure.

thanks,
bsr

grahammendick:
There are a couple of definitions of a breadcrumb trail. I've gone with the following (taken from your wikipedia link)

> A breadcrumb trail tracks and displays each page viewed by a visitor of a website...in the order the pages were viewed

This doesn't fit your purposes because it sounds like you want the other definition:

> a hierarchy of the current page in relation to the website's structure

So, yes, you will have to buid your own.

You might find it easier to use the` RefreshLink` than the `NavigationLink` for the author/book example. A `RefreshLink` is a `NavigationLink` that navigates to the current `State` so if you configure your example with one State:
```js
var stateNavigator = new StateNavigator([
  {key: 'author', route: 'author/{id}+/book/{title}'}
]);
```
then you can build a Hyperlink to the author when viewing a book like this:
```jsx
<RefreshLink
  navigationData={book: ''}
  includeCurrentData={true}>
  Author
</RefreshLink>
```

Setting `includeCurrentData` means it will keep the current author id. And clearing the book id means it'll generate the href `/author/xyz`.

grahammendick:
I should just stay that `includeCurrentData` also works with `NavigationLinks`. So if author and book are different `States`, you can build a breadcrumb from book to author like this:

```jsx
<NavigationLink
  stateKey="author"
  navigationData={book: ''}
  includeCurrentData={true}>
  Author
</NavigationLink>
```

bsr203:
thanks a ton. I think the second example with `NavigationLink` sounds more apt for me. I didn't notice `includeCurrentData`, and thanks for pointing it out. I ported almost everything from `react-router` except navigating from actionCreators through redux actions. I may move those navigation back to the components. thanks again for your help.

grahammendick:
[...]

PR #100, opened 2016-11-12
Created Navigator Component for React Native

grahammendick:
The Navigation router's crumb trail back navigation is equivalent to stack pop navigation in a native app. Combined the Navigation router with React Motion to create a native navigator, called SceneNavigator. React Motion animates the navigation when scenes change. Did this work in ReactDOM because native development is prohibitively slow on windows, but shouldn't be a problem because Navigation and React Motion both work on native.

The SceneNavigator renders a cached scene/component for each crumb in the trail and one new scene/component for the current state. Each scene is wrapped in a Motion component that animates the current state when it becomes a crumb. The array of scenes is wrapped in a TransitionMotion component that animates each scene when it's navigated to (mounts) and is navigated back from (unmounts).


bsr203:
I don't use react-native, but great to see you are actively supporting the router and thanks for all the effort. cheers.

grahammendick:
Thanks for the encouragement. Remember that discussion we had about the [breadcrumb trail](https://github.com/grahammendick/navigation/issues/92)? Turns out an auto-tracking breadcrumb trail is just the ticket on native!

Anyway, how you getting on with the router?

bsr203:
After the initial port from react-router, I didn't spent any time on router specific work.. that's a good thing as it was so unobtrusive and never met any issue. I also work on the backend, so bit slow overall, but hope to use the full potential provided by this library in the future. thanks again..

PR #142, opened 2017-08-27
Relay Modern Example

bsr203:
Sample using relay modern https://facebook.github.io/relay/docs/relay-modern.html


bsr203:
Hi. I was playing with relay v1 and copied the relay example and adapted. Everything was smooth and wrks well. Let me know any changes needed.

grahammendick:
Thanks for the great PR! It reminds me of when you first got in touch. I spent a weekend sweating over that Navigation Relay example 😅 , so I really appreciate your Relay Modern version.

I’ll get back to you with any code review comments shortly, but I wanted to send you a thank you message right away.

NavigationReact/sample/relay-modern/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "react-dom": "*",
+    "react-relay": "^1.2.0",
+    "webpack": "^3.5.5"
+  },
+  "scripts": {
+    "start": "babel-node ./server.js",
+    "updateSchema": "babel-node ./updateSchema.js",
+    "relay": "relay-compiler --src . --schema ./schema.json"
+  },
+  "devDependencies": {
```
grahammendick: I don’t use `devDependencies` in examples. I put everything under `dependencies`.

In the Relay example, `webpack` bundles at runtime on the server. This isn’t how it should be done for real but it simplifies the code. So should I put `webpack` in `devDependencies` or `dependencies`? To avoid this sort of confusion I put everything under `dependencies`.

NavigationReact/sample/relay-modern/index.js
```diff
@@ -0,0 +1,95 @@
...
+  store: new Store(new RecordSource())
+});
+
+/**
+ * Listens for navigation events and renders the Relay RootContainer, passing
+ * in the Relay Route and Component retrieved from the state.
+ */
+stateNavigator.onNavigate((oldState, state, data) => {
+  var Component = state.component;
+  var variables = {};
```
grahammendick: You don’t need the `variables` variable. You can pass the incoming `data` directly into the `variables` prop on the `QueryRenderer`.

NavigationReact/sample/relay-modern/People.js
```diff
@@ -0,0 +1,66 @@
+import React from "react";
+import { createFragmentContainer, graphql } from "react-relay";
+import { NavigationLink, RefreshLink } from "navigation-react";
+
+var People = ({ people, stateNavigator }) => {
+  var rows = people.persons.map(person =>
```
grahammendick: Can you use 4 spaces for formatting, please? (Don't bother changing the auto-generated files like `package.json` and `schema.json` - 2 spaces is fine for those).

NavigationReact/sample/relay-modern/index.js
```diff
@@ -0,0 +1,95 @@
...
+
+const modernEnvironment = new Environment({
+  network: Network.create(fetchQuery),
+  store: new Store(new RecordSource())
+});
+
+/**
+ * Listens for navigation events and renders the Relay RootContainer, passing
+ * in the Relay Route and Component retrieved from the state.
+ */
```
grahammendick: This comment relates to the original Relay example. It needs a tweak to match the approach you've taken for Relay Modern

NavigationReact/sample/relay-modern/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "navigation-relay",
+  "description": "Navigation React Relay (modern) example",
```
grahammendick: I think we should rename the package to "navigation-relay-modern" and change the description to "Navigation React Relay Modern example". What do you think?

NavigationReact/sample/relay-modern/README.md
```diff
@@ -0,0 +1,11 @@
+# Navigation Relay (Modern)
```
grahammendick: I think we should change the name to "Navigation Relay Modern". Do you think we need the brackets?

bsr203:
Hi. Thanks for the kind word and review. Was busy today and will make the changes you suggested shortly. Thanks.

bsr203:
I am not a pro with github. Please let me know anything else to do. I use prettier to format code, and hope it looks correct with 4 spaces indentation now. Let me know for anything else. Thanks.

grahammendick:
Thanks for making the changes. It’s looking great!
 
Sorry for being fussy, but can you make some more formatting changes, please? I’d like the files to be formatted the same across the Relay and Relay Modern examples. For example, I can then CTRL+Tab to switch between the `Person.js` files across the two examples and the changed lines jump out (because all the other lines match). If you’re happy, will you update the `Person.js`, `People.js` and `schema.js` so the formatting matches with the original Relay example?

bsr203:
Please see the changes and let me know.

grahammendick:
Spot on, thanks.

Why don’t we just put your example into the Relay folder and delete the original? Is that a bad idea?

bsr203:
thanks. In the future, it will be just relay-modern. relay-modern is a complete rewrite with new API, and instead of calling relay 2 (similar to Angular 2 fiasco :-) ) they opted the v1 release as the modern version with a clear update path. At this point, there is emphasis on these two version as see in the [docs](https://facebook.github.io/relay/docs/relay-modern.html) and examples. 

Edit: while I was looking at the examples, I see they are changed to only include the modern version.
https://github.com/relayjs/relay-examples/commits/master

so, it is up to you. Older example may only short time, and both example are not affected by the API of this library. so they only highlight change in relay API :-)

grahammendick:
Ok, it sounds like we should keep both versions for now.

Thanks again for such a great PR. It inspired me to get in [touch with the Relay team](https://github.com/facebook/relay/issues/2075)

bsr203:
Thank you for all the hard work and clean code base. Cheers.

PR #245, opened 2019-06-10
Moved to a Single React Root

### LINKED ISSUE #237: Native global React context out of route
If I have some global context I want to apply to all routes (login data etc.), what's the best way to do this?

### LINKED ISSUE #241: Updating context only works in a single route?
I have a global context that I need to access from all routes. If I update the context value it only takes effect in the same route. It goes back to the default value after I navigate to a different screen.

grahammendick:
Fixes #241 Fixes #237

When working on a `SearchBarIOS` component couldn’t get the search bar to show up in the navigation bar. Each screen was a separate React root. Navigating to new screen first went to native code, which pushed a new `UIViewController` and rendered the new React root into the screen content. So the screen content was loaded **after** the new view controller was pushed. Because the search bar wasn’t rendered before the new view controller transitioned it didn’t show up.

@jacobp100 came to the rescue. His [NavigatorTest](https://github.com/jacobp100/NavigatorTest) project showed how to have a single React root for all screens, instead of one for each screen. With a single React root the timing problem disappears because the new screen is rendered in React **before** it’s added on the native side. Native pushes the new view controller and immediately sets its content from the added screen.

Although the search bar was the driver for the change, there are many other benefits.

1. Don’t get a blank screen during a transition while waiting for the screen to render
2. Screens share React context because they share a React root
3. It’s all component and props. For example, there’s a `TabBarIOS` component for `UITabBarController` and `unmountStyle` prop for custom Android animation
4. Requires no native setup so works on Expo if they can be persuaded to include it
5. Don’t have to wait for transition to complete before iOS bar buttons appear


jacobp100:
This is awesome! 😀

grahammendick:
Giant thank you for your NavigatorTest project and for pushing me in the right direction

PR #263, opened 2019-07-12
Add barTintColor and tintColor to NavBarIOS and TabBarIOS

mpiannucci:
More styles could be added in the same way down the road.


mpiannucci:
I just realized I need to expose the unselectedBarTintColor for TabBarIOS so I’ll do that tomorrow

mpiannucci:
The one last thing is that the style props are now applied in `viewWillAppear` but that gives a super rigid transition. I think using `willMoveToSuperView` might give a better transition.

grahammendick:
Hey, thanks for putting the PR together! Is it ready for review because I couldn't get it to build? I think you mean `_tabBarController` instead of `self.tabBarController`. And there's not setter for the `isTranslucent` property on `tabBar`. Sorry if I looked at it too soon. Thanks again

mpiannucci:
Sorry about that, I think I forgot to push my last updates. I’ll fix those and get back to you after

mpiannucci:
Okay should be all set now! Let me know what you think, its all pretty basic

mpiannucci:
This causes #265

grahammendick:
Nice catch! I'll hold off reviewing this until you're happy

mpiannucci:
Okay sorry about all the issues, this should be good to go now. I got overzealous by including the translucency settings becasue they change the render surface. For now, I'll just leave them as default.

grahammendick:
Thanks so much for the PR. Please don’t be put off by my long review comments. I like to give quite a bit of detail to help us arrive at a shared understanding of the feature.

## Resetting Defaults
These `tintColors` are tricky because they live on the `navigationBar` on the `UINavigationController` and not on the `navigationItem` on the `.UIViewController`  I think this is a design flaw of the iOS platform. It causes problems because properties on the `navigationBar` apply to all screens, unlike properties on the `navigationItem` which are isolated to the screen you’re in. 

Take the example of a navigation from A → B. Let’s say that Screen A doesn’t set the `barTintColor` and that Screen B sets it to green. When Screen A loads it has the default color but after navigating to Screen B and then back, Screen A has the color green! That’s because the settings from B were made to the shared `navigationBar`.

To fix this, we must set the default `barTintColor` if the prop is null. We must do that in both the `NVSceneController` as well as in `NVNavigationBarView`.  

## Text Attributes
It was interesting to see you set the color of the titles using the TextAttributes (I never knew about them before). There are quite a lot of TextAttribues, not just color. I don’t think we want to introduce new props for each of them. I had a look at how React Native handles them for its `Text` component. It uses the `style` prop instead of separate props for each of them. Couldn’t we do the same?
```jsx
<NavigationBarIOS barTintColor=”green” style={{color: 'white'}} />
```
Then the rest of the TextAttributes become extra style props
```jsx
<NavigationBarIOS barTintColor=”green” style={{color: 'white', fontSize: 20}} />
```
I got a bit lost in the React Native code. Can you work out how it uses the [TextAttributes class](https://github.com/facebook/react-native/blob/master/ReactCommon/fabric/attributedstring/TextAttributes.h)? Again, we must reset the defaults because you can set the text color and then clear it.

mpiannucci:
# Resetting Defaults

Agreed this is important. Another part of it is that usually an app will want to have a default `barTintColor` for the app which can be overridden in each Screen. That also means there should be default `style` for the `title` as well most likely? What do you think? Cuz the hardest part of that is just making sure there are known defaults to revert to when navigating away from a screen. 

# Text Attributes

So I looked into this a bit. In addition to the `cpp` version you linked, there is alos [the objective c version](https://github.com/facebook/react-native/blob/master/Libraries/Text/RCTTextAttributes.h). Then if you trace back the `Text` view you get to the shadow view where the attributes are actually computed and applied [here](https://github.com/facebook/react-native/blob/4037f08b72866003988b417c50b0431193c98175/Libraries/Text/BaseText/RCTBaseTextShadowView.h#L21). They are exposed to JavaScript using the [BaseViewManager](https://github.com/facebook/react-native/blob/e4186e92ddbad927899579ad29962f19af5e485b/Libraries/Text/BaseText/RCTBaseTextViewManager.m#L29). 

So I think we can do something just like this but just simpler. Ill play around with it when I get a chance.

grahammendick:
Won't setting `barTintColor` to `nil` reset to the default?

mpiannucci:
It might, I have never tried. I was also saying that you could expose a wayt o customize what the default is so you dont have to set the color in every single screen unless you want to override it. Does that make sense?

grahammendick:
I'm not keen on global defaults because it seems to go against the React grain. I think we should stick to just components and props and leave it up to  the user to devise their own approach to theming. For example, they could add their Theme to React Context
```jsx
const theme = useContext(ThemeContext);
<NavigationBarIOS style={{color: theme.color} />
```

mpiannucci:
Fair enough! Ill play with the text styles soon

mpiannucci:
The `style` props work totally fine as expected. The hard part now is resetting to defualts. I tried to get this working but ht a problem where the props for the navbar are not set again when navigating back. Moving forward works fine becasue everything comes in the correct order so you can set the props, then set the nav bar property members to nil when the view navigates away, then the new props are passed down for the next view. But when you click the back button, the original navbar props are not passed again, they are expected to have persisted in the NavBar.

grahammendick:
When you navigate back from B to A, iOS calls `viewWillAppear` on Screen A. This is where you reset the navigation bar style from the `NVNavigatonBarView` properties. I think it's enough to remove the null checks from `viewWillAppear`.

mpiannucci:
Coulda sworn that was the first thing I tried. It works fine but I have to find a way to make the transition between bar colors more smooth when navigating back.

grahammendick:
Have you tried `viewDidAppear` instead of `viewWillAppear`?

mpiannucci:
Yeah no change really. Its not terrible, but also not super smooth.

grahammendick:
Have you seen the [NavBarColorAnimation example](https://github.com/kamwysoc/NavBarColorAnimation)? I haven't tried it out but it claims to smoothly animate bar tint color changes. If navigating from A → B and going from red to blue, the trick is to set the [bar tint color to red inside B's willMoveToParent](https://github.com/kamwysoc/NavBarColorAnimation/blob/0b579448db996f49ca16d01326cf84ae793dc6ad/navAnimation/ViewControllerB.swift#L17).

grahammendick:
Even if we can smoothly animate the color, what about the other styles like font and text color? Will the same approach work for them? Does it matter?  What is the normal usecase for changing the navigation bar? Do people usually do it system wide or on a screen by screen basis?

mpiannucci:
Yeah the same approach works fine for font and other styles. The text color prop is working with the current approach. I am planning to add more style options once I am happy with the animation.

In my experience, apps tend to theme the navbar and tab bar, etc once using UIAppearance hooks so they have a consistent appearance through the app. But the ability to have screen by screen is still an important use case. 

Even further, it’s common to place custom views in the navigation item title view, so that’s a future use case as well.

I saw the same stack overflow post. My thought is maybe it’s possible to use willMoveToSuperview to grab the next screens style props and then set them.

grahammendick:
Thanks. Sorry for firing questions at you!

That stackoverflow implies that it should animate smoothly after iOS 10.3. Is that true?  If that is true then maybe we don't need to work on it for the first release?  If it's not true then that raises its priority I think. 

Custom views in the title view should sounds like a nice feature.  It could work something like this?
```jsx
<NavigationBarIOS>
  <TitleIOS>
    <View><Text>Home</Text></View>
  </TitleIOS>
</NavigationBarIOS>
```

grahammendick:
I don't think we should put too much effort into working around 'bugs' in the iOS platform. If the problem exists on iOS apps built without React Native then I'm OK with it.

grahammendick:
Wha tI mean is if you change the navigation bar color in an iOS app (without React Native) and the animation isn't smooth, then there's no need for us to make it smooth in React Native

mpiannucci:
> That stackoverflow implies that it should animate smoothly after iOS 10.3. Is that true? If that is true then maybe we don't need to work on it for the first release? If it's not true then that raises its priority I think.

Ill upload a gif either tonight or this weekend with how it looks and you can see how it currently looks. 

> What I mean is if you change the navigation bar color in an iOS app (without React Native) and the animation isn't smooth, then there's no need for us to make it smooth in React Native

Agreed. TBH I cannot find a single app on my phone that ever has a Navbar change from one solid color to another when navigating with a `push`. 

> Custom views in the title view should sounds like a nice feature. It could work something like this?

Yeah I think that would be cool. A lot of iOS apps use this feature so it could be useful.

mpiannucci:
Sorry for the delay... heres a video showing the current status of the behavior. As you can the problem is only with the top part of the navbar when popping back in the stack. 

ezgif com-video-to-gif

grahammendick:
Thanks. Looks good to me because that's how iOS behaves. 
How does `viewDidAppear` compare with `viewWillAppear`?

mpiannucci:
`viewDidAppear` triggered the switch later so the transition was worse to my eyes

grahammendick:
Cool. Happy with that, thanks

grahammendick:
How you getting on?

mpiannucci:
Hey sorry, busy week with work and another side project needing to be attended to. I’m gunna try to add a bunch of style options tomorrow and you can review from there?

mpiannucci:
Sorry a little backed up this weekend. Promise I will finish this up soon

mpiannucci:
Okay so the color options are definitely ready now. But when it comes to font it gets tricky, because if there are BarButtons added they will not have the same font applied by defualt. So if we expose this in the `style` of the navbar it might be misleading unless the font options are prefixed with `title` which might be non standard. Its up to you what you want to do, but as a first step the colors are ready.

grahammendick:
Great point about the font and bar buttons. Perhaps this is a good time to take a step back and have a think about what the props should be and what affect they should have.

For example, I'm not sure anymore that `barTintColor` is the same thing as `backgroundColor`. I think they're only the same if the translucence is false. Same goes for `tintColor` and `color`.

Also, I'm not keen on one prop being used to set multiple iOS properties like we're doing with color. Can tintColor and the text foreground color be different? If so, why would they be different?

Should we just have a `tintColor` prop that just sets the `tintColor` property of the navigation bar? And the same for `barTintColor`?
We could leave the tricky decisions about the style prop and the text attributes to a different PR. Or do you we need to think about them now?

grahammendick:
I think we've got three options
```jsx
<NavigationBarIOS tintColor="blue" />
<NavigationBarIOS style={{tintColor: "blue"}} />
<NavigationBarIOS style={{color: "blue"}} />
```
Which one do you prefer? Which one is most consistent with the way React Native does things? Which one works best with the other style/props we need to add?

grahammendick:
I think we can address your point about fonts and bar buttons in this PR. We don't want styles to "cascade" because that's not how React Native styles work. So the tintColor of the NavigationBar shouldn't apply to the BarButtons. If we add a tintColor prop to the BarButtonIOS component then it shouldn't inherit the one from the NavigationBar. Have I understood the problem?
```jsx
<NavigationBarIOS barTintColor="whilte" style={{tintColor: "blue"}} />
<BarButtonIOS style={{tintColor: "green"}} />
```

grahammendick:
Alright, I've had a play around and I understand why you're setting multiple iOS properties from one prop. The `tintColor` changes the back button text and the bar button text but not the title text. To change the title text you need the title text attributes. 

I'm still not clear about how we should manage the style/props, though. The problem is that styles don't cascade in React Native. But if we set a tintColor on the navigation bar it 'cascades' to the bar buttons. I thought we could get around it by adding a tintColor to the bar button but if the user sets the tintColor for the navigation bar to red and doesn't set the tintColor on the BarButton then how do we make sure that the bar button is the default color? If we set it to nil it will end up red.

grahammendick:
How about this for the props and style?
```jsx
<NavigationBarIOS barTintColor="blue" tintColor="white" titleStyle={{color: 'white'}} />
```
Calling it `titleStyle` makes it clear they apply to the title. It gives us scope for adding more title styles like `font` and `fontSize`.

mpiannucci:
Okay at first glance that looks good. So tint would apply to the buttons only and the title would be separate? 

That makes sense, but I do want to be sure you can override it in a per button basis probably right?

grahammendick:
Does it make sense to override it per button? When would the right bar button be a different color to the Back Button? 

I couldn't work out how to change the Back Button text. Do you know how?

grahammendick:
Take a look at the props of [NavigatorIOS](https://facebook.github.io/react-native/docs/navigatorios.html#props). It was the first ever navigator, back when React Native only supported iOS. It had props for `tintColor`, `barTintColor` and `titleTextColor`. Perhaps we should copy that?

grahammendick:
I think these three props are all we should allow
```jsx
<NavigationBarIOS barTintColor="white" tintColor="blue" titleColor="blue" />
```
Any more props don’t make sense. To see why, let’s say we want a prop that changes the font of the bar buttons. There’s  no point changing the font of the right bar button if we can’t also change the font of the back button. It would look odd if they had different fonts. However, the only way to change the font of the back button is to create a new UIButton in the previous screen. There’s no way to change the font of the back button inside the current screen. So there’s no point adding a prop to change the font of the bar buttons because we can’t add a prop that changes the font of the back button. If we want to change the font of bar buttons we can do it the native way by changing the UIAppearance. 

What do you think?

mpiannucci:
I agree with you. Those three props are the most important ones for the navbar.

It also doesnt make the API such that if we want to allow custom title views or fonts in the future it wouldn’t break anything.

grahammendick:
Great. Let's go with that then. You OK to make the changes?

mpiannucci:
Yep will do this weekend, should be straightforward

mpiannucci:
Okay this should be all set now with the new prop names. I also used the same type of prop names for the tab bar to match.

grahammendick:
Thanks. I'm knee-deep in a fix at the minute but looking forward to working on your cool PR next

NavigationReactNative/sample/twitter/App.js
```diff
@@ -26,7 +26,7 @@ stateNavigator.navigate('home');
 
 export default () => (
   Platform.OS === 'ios' ? (
-    <TabBarIOS>
+    <TabBarIOS barTintColor="#ffffff" tintColor="#006dbf">
```
grahammendick: I changed the `tintColor` but it didn't seem to do anything. Is it meant to change the color of the tab titles? Don't we need the same approach as you took with the `NavigationBar` and set the `titleTextAttributes`?

NavigationReactNative/sample/zoom/Detail.js
```diff
@@ -7,7 +7,7 @@ export default ({colors, color}) => (
   <NavigationContext.Consumer>
     {({stateNavigator}) => (
       <ScrollView contentInsetAdjustmentBehavior="automatic">
-          <NavigationBarIOS title="Color">
+          <NavigationBarIOS tintColor={color} titleColor={color} title="Color">
```
grahammendick: What an ingenious way to demo the feature! Love it!

NavigationReactNative/src/ios/NVNavigationBarView.m
```diff
@@ -19,6 +19,21 @@ - (void)didSetProps:(NSArray<NSString *> *)changedProps
     if ([changedProps containsObject:@"title"]) {
         [self.reactViewController.navigationItem setTitle:self.title];
     }
+    if ([changedProps containsObject:@"barTintColor"]) {
```
grahammendick: We don't need to check if the props have changed. We can always set the colors even if they haven't changed. I needed to check for the `title` because the `title` can be set in two different places. If it isn't set by the `NavigationBar` then I don't want to overwrite any existing value. We don't have that problem with colors because they're only set by the `NavigationBar`.

build/npm/navigation-react-native/navigation.react.native.d.ts
```diff
@@ -196,7 +208,20 @@ export class TabBarItemIOS extends Component<TabBarItemIOSProps> {}
  * Defines the Tab Bar Props contract
  */
 export interface TabBarIOSProps {
-    children: React.ReactElement<TabBarItemIOS> | React.ReactElement<TabBarItemIOS>[]
+    children: React.ReactElement<TabBarItemIOS> | React.ReactElement<TabBarItemIOS>[];
```
grahammendick: Thanks for adding the missing semi-colon!

NavigationReactNative/src/ios/NVNavigationBarView.m
```diff
@@ -19,6 +19,21 @@ - (void)didSetProps:(NSArray<NSString *> *)changedProps
...
+        [self.reactViewController.navigationController.navigationBar setBarTintColor:self.barTintColor];
+    }
+    if ([changedProps containsObject:@"tintColor"]) {
+        [self.reactViewController.navigationController.navigationBar setTintColor: self.tintColor];
+    }
+    if ([changedProps containsObject:@"titleColor"]) {
+        [self.reactViewController.navigationController.navigationBar setTitleTextAttributes:
+            @{NSForegroundColorAttributeName:self.titleColor}];
```
grahammendick: This throws an error if `titleColor` is nil. You can try this out by setting the prop as below and then selecting red followed by orange when on the details screen.
```jsx
titleColor={color === "red" ? color : undefined}
```
You took a different approach on the `NVSceneController`  by setting the `titleTextAttributes` to nil. Although this works it's a bit of a sledgehammer because it will clear out all attributes that the user has set. For example, they might have set the title font using `UIAppearance` (something we want to support because it's hard to do with props alone). 

We can get a mutable copy of the `titleTextAttributes`, just remove our one attribute and then set it back. Whatever approach you take check that it works with `UIAppearance`.
```objective-c
NSMutableDictionary *titleAttributes = [self.reactViewController.navigationController.navigationBar.titleTextAttributes mutableCopy];
[titleAttributes removeObjectForKey:NSForegroundColorAttributeName];
if (self.titleColor != nil) {
    titleAttributes[NSForegroundColorAttributeName] = self.titleColor;
}
[self.reactViewController.navigationController.navigationBar setTitleTextAttributes:titleAttributes];
```

NavigationReactNative/sample/twitter/App.js
mpiannucci: Whoops this should have been updated. I changed the name of the property to `selectedTintColor` to be more clear.

NavigationReactNative/src/ios/NVNavigationBarView.m
mpiannucci: yeah thats a better way to do it for sure. I will update and test that out thanks

mpiannucci:
Made the requested changes. Let me know if you have any more feedback

mpiannucci:
Sorry closed by accident

mpiannucci:
Thanks for catching all of that!! Haha I work on this randomly after my day job so sometimes my brain is always fully awake!!

mpiannucci:
I think the typings should already be up to date? 

Got rid of the project files though!

grahammendick:
Thanks! You're right about the build typings. The typings file you need to copy them to sits inside the [test folder](https://github.com/grahammendick/navigation/blob/master/NavigationReactNative/test/node_modules/%40types/navigation-react-native.d.ts)

Also, if you have any suggestions about how I can improve the contribution process then please let me know.

mpiannucci:
Gotcha no problem!!!

Its been a pretty easy process!! I feel bad I have been a little sloppy along the way but pumped to be using this project

grahammendick:
Good to hear. I’ve enjoyed working with you. Hopefully we’ll work together on another PR some time soon.

grahammendick:
No biggie, but I just realised that, instead of `didSetProps` in `NVTabBarView`, we should’ve used setters
```objective-c
- (void)setBarTintColor:(NSString *)barTintColor
{
   [_tabBarController.tabBar setBarTintColor:self.barTintColor];
}
```
Can't do this in `NVNvigationBarView` because it needs access to the `reactViewController`

grahammendick:
Spotted a small bug with the Zoom example. If you select 'red', for example, then press the back button then the title on the Grid screen stays red. It doesn't happen with the large title and only happens if you press back (swiping or pressing cancel works fine). Have you got a real iPhone that you can test this on because it might just be a bug with Xcode?

mpiannucci:
Yeah ill check it out later tonight... that is really weird that the back button would have different behavior than swiping.

grahammendick:
Thanks. I'm not too fussed because I don't think it's a realistic use-case. Perhaps I'll just remove the navigation bar colors  from the zoom example?

mpiannucci:
That is a crazy bug. I really want to fix it/find out why it happens. The same thing happens when large title is enabled, the color only shows when you scroll but the color of the small title is not reset when the actual back button is clicked but any other way and it is fine. 

Do you have any idea if different methods are being called by the navigation router depending on swipe vs back button?

mpiannucci:
by setting the colors in viewdidappear the problem seems to go away

mpiannucci:
Going to submit another pull request with a very simple fix

Issue #281, opened 2019-08-14
Additional features to TabBarIOS/TabBarItemIOS


arunmenon1975:
The app that i am developing has Tabs as the primary navigation artefact. I am using 
 `TabBarIOS` and `TabBarItemIOS` from the `Navigation React Native` router. Everything works fine from the tab functionality perspective but for two issues, one of which is UX related and the other is more of a project convenience than a requirement.
 
1) How do i add badges to the tab items (a big UX plus and also present in my wireframes). In addition, and it just struck me now, how do i style the tabitems/tabbar? The style prop doesn't seem to work.
2) Use of vector icons in place of images for the icons 
 
While definitely not a showstopper since functionally everything works fine, I decided to roll-up my sleeves and try and modify the sources to see if i could to get at least the badge going(AFAIK even React Native Navigation uses images and/or i may need to go the `getImageSource` route) . I have to dip my toes at some point and this seemed a good enough usecase.

For issue 1, I made the following changes but i seem to be missing something since it just doesn't seem to work (there are no errors, however):

My tab file:

```
<TabBarItemIOS  badgeColor="red" badge="123" title="Menu">
       ......
</TabBarItemIOS>
```

In 'navigation.react.native.d.ts':
   ```
...
    badge?: string;
    badgeColor?: string;
...
```

In 'navigation.react.native.js':
 ```
 ...
    var title = _a.title, badge=_a.badge, badgeColor=_a.badgeColor, image = _a.image, systemItem = _a.systemItem, onPress = _a.onPress, children = _a.children;
    console.log({ title: title, image: image, badge: badge, badgeColor: badgeColor, systemItem: systemItem,}) // everything except title and onPress is undefined
    return (React.createElement(NVTabBarItem, { title: title, image: image, badge: badge, badgeColor: badgeColor, systemItem: systemItem, onPress: function (event) {
            event.stopPropagation();
            if (onPress)
                onPress(event);
        } }, children));
...
```

In 'NVTabBarItemManager.m':
 ```
 ...
    RCT_EXPORT_VIEW_PROPERTY(badge, NSString)
    RCT_EXPORT_VIEW_PROPERTY(badgeValue, UIColor)
...
```

In 'NVTabBarItemView.m':
 ```
 ...
- (void)setBadge:(NSString *)badge
{
    self.tab.badgeValue = badge;
}
- (void)setBadgeColor:(UIColor *)badgeColor
{
    self.tab.badgeColor = badgeColor;
}
...
``` 

[...]

PR #284, opened 2019-08-16
Added badge and badgeColor to TabBarItemIOS

arunmenon1975:
In reference to #281, added new props `badge` and `badgeColor` to `TabBarItemIOS `.


grahammendick:
Thanks for the PR. You’ve done a great job.

Have you tried setting a `badge` and a `systemItem`?
```jsx
<TabBarItemIOS badge=”3” systemItem="contacts">
  ...
</TabBarItemIOS>
```
 It probably doesn’t work because `systemItems` are weird. Take a look at the `setSystemItem` in `NVTabBarItemView` and you’ll see that I had to create new `UITabBarItem` because there’s no `systemItem` property. That means that the badge (and badgeColor) will get overwritten unless we restore them.
```objective-c
- (void)setSystemItem:(UITabBarSystemItem)systemItem
{
    NSString *currentBadge = tab.badgeValue;
    self.tab = [[UITabBarItem alloc] initWithTabBarSystemItem:systemItem tag:0];
    [self setBadge:currentBadge];
}
```    
Make sure you can recreate the problem before you fix it.

 We should include the badge in the [Twitter example](https://github.com/grahammendick/navigation/tree/master/NavigationReactNative/sample/twitter). The Twitter example has a Notifications tab with three items so we could add a badge value of "3" to it. Then, when the notifications tab is clicked, we can remove the badge value. Is this how badges work on iOS?

arunmenon1975:
I just tried that on my currently in-dev app  which has the following set:  

```
<TabBarIOS 
        selectedTintColor="orange" 
        unselectedTintColor="yellow" 
        barTintColor="gray">
           <TabBarItemIOS 
                onPress={()=>this.setState({title:"Menu"})} 
                 systemItem="contacts"
                  badge="123" 
                   badgeColor="green"  
                   title="Home"
                 >
                      ....
               </TabBarItemIOS>
                <TabBarItemIOS  
                    title="Daily Specials" 
                    systemItem="favorites"
                     onPress={() => {
                       if (!stateNavigatorDailySpecials.stateContext.state)
                            stateNavigatorDailySpecials.navigate('dailyspecials');
                            this.setState({title:"Daily Specials"})
                       }}> 
                      ....
               </TabBarItemIOS>

</TabBarIOS>
```
and renders:

Screenshot 2019-08-16 at 7 26 58 PM



The problem you mentioned doesn't seem to reflect in my checks, but what i do notice is that my custom `title`  prop is getting overwritten with the system-set label. I suspect that this may not be exactly ideal for everybody, including my own current setup if i do opt to use `sytemItem`. 

Screenshot 2019-08-16 at 7 05 27 PM


Out of curiosity, I also attempted adding an image prop to see what gets rendered:

`image={require("../../resources/media/images/icons/menu-menu.png")`

but it looks like the `systemItem` prop takes precedence(similar to the second screenshot i shared).

I am on a Simulator and not an actual device. I am not sure if the problem you mention exists in other devices but i have my test device as `iPhoneXr` on `i0S 12.4`.

Regarding the integration into the Twitter sample, i will have a go at it.  The last time i used an actual iPhone was way back in 2013-14  so cant exactly recall how the badges on tab items worked out of the box. But the scenario you suggested seems to be the expected behavior i guess, at least from a user expectation perspective. It also provides for an interesting and usable twist to the sample so perhaps it can be added in without too much thought into actual/guidelined behavior. Its a sample after all and can be turned off. Or at least thats my opinion.

grahammendick:
We have to recreate the problem before we can fix it. This time, instead of setting the `systemItem` prop straight away, only set it after a `setState`. I’m confident that this will remove the `badge` value. 

```jsx
<TabBarItemIOS
  badge="i23"
  title={!this.state.icon ? "Home" : undefined}
  systemItem={this.state.icon ? "contacts" : undefined}
  onPress={() => this.setState({icon: true})}>
  ...
</TabBarItemIOS>
```
I don’t think iOS supports a `systemItem` with a custom title. Likewise, once you have a `systemItem` then you can’t set an image.

grahammendick:
Just checked and setting a `systemItem` prop from a `setState` doesn't work anyway! My bad. Let's not worry about it for this PR, then.

Let's focus on adding a badge to the Twitter example.

arunmenon1975:
> Just checked and setting a `systemItem` prop from a `setState` doesn't work anyway! My bad. Let's not worry about it for this PR, then.
> 
> Let's focus on adding a badge to the Twitter example.

ok.

Since the Twitter example showcases the possibilities in the library i guess i cannot go overboard and attempt to reach a solution just for the sake of reaching it; have to  have it vetted. 

Before i make changes i just wanted to check if my thought process is ok:

- The Notifications tab data is currently retrieved from the `getFollows()` method which maps over a hardcoded `notifications` array and then retrieves the id from the `account` object based off the values. The original reason to hardcode the number and value of items in the array was more from the convenience in getting the data and populating the screen rather than accounting for near realtime usage scenarios. 
- The badge information might be a server-driven value or perhaps uses device storage 
- The badge information needs to be available when the app opens up (currently there is no subscription to the `stateNavigator` context as the tab is declared prior to the Navigation stack declaration. This would mean using simple in-component state just to showcase the dynamic change rather than pass data via `Navigator` methods.

My rough outline:

In data.js:

  ```
const userBehavior = {
    hasCheckedNotifications:false
  }
```
_[The nearest simulation of data that might be coming from the server or after reading device storage values]_

  ```
const getHasCheckedNotifications=()=>{
    return userBehavior.hasCheckedNotifications
  }
```
_[Call the above method to populate the data]_

  ```
const setHasCheckedNotifications=(bool)=>{
     userBehavior.hasCheckedNotifications=bool;
  }
```
_[Call the above method to set the status maybe after a `onNavigate` or `onBeforeNavigate`]_

`  export {getHome, getFollows, getTweet, getTimeline, getHasCheckedNotifications,setHasCheckedNotifications };
`
_[Expand the named exports]_

In App.js:

```
notificationsNavigator.onNavigate((oldState, state,data, asyncData, stateContext)=>{
  let hasChecked = getHasCheckedNotifications();
  if(!hasChecked){
    setHasCheckedNotifications(true)
  }
})
```
_[Have a listener that reads and sets the data if required]_


```
<TabBarItemIOS 
      badge={!notificationsNavigator.stateContext.state?getFollows().length:null} 
      badgeColor={!notificationsNavigator.stateContext.state?null:"blue"} 
      title="Notifications" onPress={() => {
        if (!notificationsNavigator.stateContext.state)
          notificationsNavigator.navigate('notifications');
      }}>
```
_[conditionally set the badge prop from reading the length of the notifications rather than hardcode]_

With these WIP changes i get the initial badge:

Screenshot 2019-08-16 at 11 06 21 PM

I am able to reset the `hasCheckedNotifications` property when i navigate to the Notifications tab but of course it will not reflect since the value is not bound. I am wondering how best this can be done. Some options:

- use local state, simple but perhaps not really showcasing Navigator's capabilities
- `redux` etc or `React context` or other state management solutions.. perhaps overkill for an example that is not purely focussed on the badging capability
- device storage read/write perhaps via `asyncStorage`..similar to above, perhaps an overkill

Do let me know if you have any preferred approach since the example showcases several other capabilities and i don't want to make unnecessary changes that effect other functionality

**Comment by grahammendick** [comment]
Thanks for writing down your plan. Let’s keep it simple. We’re only showing off your badge PR, remember!

So, your idea of local React state works for me. Add a `checkedNotifications` to the state of the App component and default it to false. When the user clicks the notifications tab then set it to true. Then set the badge from the `getFollows().length` if notifications haven’t been checked yet.
```jsx
<TabBarItemIOS
    badge={!this.state.checkedNotifcations ? getFollows().length : undefined}
```
I don’t think we should set a badge color. It’s more obvious that we’re using the underlying native iOS API if we use the default color.

Also, it's up to you if you want call `setState` or `useState`.

grahammendick:
So no need to change data.js or add an `onNavigate`. But thanks again for working so hard on this

arunmenon1975:
> So no need to change data.js or add an `onNavigate`. But thanks again for working so hard on this

I have updated the PR. I half thought of opening a new branch but persisted with this since they sort of seem related. Please see https://github.com/grahammendick/navigation/pull/284/commits/5dec4bfbfda2eef6a95820bf76fdc4dc349bdae3

NavigationReactNative/sample/twitter/App.js
```diff
@@ -24,29 +24,40 @@ timeline.renderScene = ({id}) => <Timeline timeline={getTimeline(id)}  />;
...
-    <NavigationHandler stateNavigator={stateNavigator}>
-      <NavigationStack
-        crumbStyle={from => from ? 'scale_in' : 'scale_out'}
-        unmountStyle={from => from ? 'slide_in' : 'slide_out'}>
-      </NavigationStack>
-    </NavigationHandler>
-  )
-);
+export default() => {
+  const [isNotified,
```
grahammendick: Good name choice. Maybe we can go one better and change it to just "notified"?

NavigationReactNative/sample/twitter/App.js
```diff
@@ -24,29 +24,40 @@ timeline.renderScene = ({id}) => <Timeline timeline={getTimeline(id)}  />;
...
+          <NavigationHandler stateNavigator={stateNavigator}>
+            <NavigationStack/>
+          </NavigationHandler>
+        </TabBarItemIOS>
+        <TabBarItemIOS
+          title="Notifications"
+          badge={!isNotified?getFollows().length:null} 
+          /* badgeColor="red" */
+          onPress={() => {
+            setIsNotified(!isNotified?!isNotified:isNotified);
```
grahammendick: You can abbreviate this to `setIsNotified(!isNotifed)`

NavigationReactNative/sample/twitter/App.js
arunmenon1975: Ok. Will make the change.

NavigationReactNative/sample/twitter/App.js
arunmenon1975: Ok. Will make the change.

grahammendick:
I noticed you made quite a few formatting changes. Did you mean to or did prettier do it?

arunmenon1975:
> I noticed you made quite a few formatting changes. Did you mean to or did prettier do it?

In short, I meant to but did hesitate for about a second.

I use an IDE plugin called `[react-beautify](https://marketplace.visualstudio.com/items?itemName=taichi.react-beautify)` which is currently set to `prettydiff` as the default formatter implementation. I am guessing it is `prettier` with `React` support. 

I was in two minds since i could see a large number of  LoC changes prior to committing but i went ahead since the file in question is not a 'core' file (in this particular context, where files within the native directories take precedence along with other ts/js files that are required for the actual functioning). 

I guess its not a best-practice since people who are just casually browsing through the code changes will probably assume that there are more changes than what did actually change, which is perhaps about 3-4 lines. Introduction of the curly braces and the explicit `return` sort of influenced the decision to do a block-level format. Incidentally, I also ended up fine-tuning a couple of formatting changes that i wished would have happened automatically. 

I can revert back since i guess future changes by anybody to the file using a different formatter will again introduce the large count in LoC changes.

grahammendick:
Thanks for the detailed explanation. Don't worry about it this time but best to turn the plugin off next time (I hope we'll work together on another PR soon)

NavigationReactNative/sample/twitter/App.js
arunmenon1975: `setIsNotified(!notified)` wont work as expected if the user switches tabs (i.e. taps on Notifications) every other round since the value will cycle through `true` and `false` with each tap.

`setIsNotified(!notified?!notified:notified)` will work however. Logically however `setIsNotified(!notified)` is correct but because it is simple state implementation for a visual experience, this quirk is introduced.

grahammendick: Good point. Should just be `setIsNotified(true)`

arunmenon1975: That should have been obvious. Will make that change.

arunmenon1975:
> It's looking great, thanks!
> 
> The only other thing to do is to update the [TypeScript typings for the TabBarItemIOS component](https://github.com/grahammendick/navigation/blob/f8d8bde3d21d4d504105c8866bfee508ba83c651/NavigationReactNative/test/node_modules/%40types/navigation-react-native.d.ts#L178) with the `badge` and `badgeColor` props. These changes need to be [copied over to the typings in the build folder](https://github.com/grahammendick/navigation/blob/master/build/npm/navigation-react-native/navigation.react.native.d.ts)

I had a doubt regarding this change. The update mentioned 

> The only other thing to do is to update the TypeScript typings for the TabBarItemIOS component...

 was made. My IDE marks it as a changed file for Git to process. So when i `commit`, this change should go in, which is expected. 

I did a `npm run package` after this change so that the `build` directory is updated.

You also mention  
> These changes need to be copied over to the typings in the build folder..

I copied over the changes to the file. The `.gitignore` file has the `build` directory listed. Should this file(where i copied the changes) also be committed? Or is it just for my local reference?

arunmenon1975:
> Thanks for the detailed explanation. Don't worry about it this time but best to turn the plugin off next time (I hope we'll work together on another PR soon)

I guess my inexperience in the otherwise smooth 'Github PR merging' process is shining through. Thanks for the patience. Regarding future PRs, i have noticed - from simultaneous development of the app - a few opportunities where i feel that a few modifications or additions could be considered for merging. Ultimately, a favourable consideration may depend on the principles that the library abides by or decides to abide by.

grahammendick:
The typings in the build folder should be committed too. The `.gitignore` doesn't list the build folder so all should be good. The `npm run package` doesn't copy over the typings. It's a manual process to apply them in the two places and then check in both files

grahammendick:
> Regarding future PRs, i have noticed - from simultaneous development of the app - a few opportunities where i feel that a few modifications or additions could be considered for merging.

Can't wait to hear your ideas. The more the merrier!

arunmenon1975:
> ..It's a manual process to apply them in the two places and then check in both files

I seem to have 4 changes after my commit(not pushed yet):

```
#	modified:   NavigationReactNative/sample/twitter/App.js
#	modified:   NavigationReactNative/test/node_modules/@types/navigation-react-native.d.ts
#	modified:   build/npm/navigation-react-native/navigation.react.native.d.ts
#	modified:   package-lock.json
```

The first 3 are valid including the `setIsNotified(true)` change. I am not sure about the changes to the lock file. I am assuming that since i had to do a `npm install` prior to 'npm run package' (since the command couldn't find `Gulp`, and possibly other dependencies as well) the lock file was either created or modified. 

My previous commit was the one that i pushed. I am thinking about an 'uncommit' equivalent so that i can roll back, perhaps with `git reset --soft HEAD^` or `git reset --hard HEAD^`, and then deleting the `package-lock.json` file. I can then continue with a fresh `commit` so that just the 3 files get checked in. Im not sure if this is an ok approach.

Sorry for all these minor questions. A few more hands-on attempts and things will get ironed out.

grahammendick:
Sounds like a good plan. I prefer it if the `package-lock.json` doesn't appear.

Are you getting errors when you run `npm run package`? You don't need to install `gulp` globally. There shouldn't be any global dependencies

arunmenon1975:
I initially got errors ('couldn't find gulp'  or something like that was there for sure. Didn't read through all the other errors, if any)  when running `npm run package` since i obviously didn't have any of the dependencies installed. So i installed them. Let me try with the 'Unstage Changes' and will update here.

arunmenon1975:
Please see the changes pushed to https://github.com/grahammendick/navigation/pull/284/commits/eea67f7ba5d86e9b2098e53a579f0f2762ec80e8

grahammendick:
Looks awesome! One last tweak, sorry. It should be "setNotified" instead of "setIsNotified". The confusion is my fault because I used "setIsNotified" when commenting on a different review comment. 
```js
  const [notified, setNotified] = useState(false);
```

arunmenon1975:
Ok. Will make the change, push and also update here as well.

arunmenon1975:
Pushed the changes. Please see at https://github.com/grahammendick/navigation/pull/284/commits/f4648dd25cecaccfad8085089a7e334232e354c6

grahammendick:
Great. Leave it with me to play with but all looks very good!

grahammendick:
Thanks, it works a treat!  I'll make a few formatting tweaks but other than that it's an awesome bit of work.

PR #285, opened 2019-08-16
Fix iOS navbar title color not resetting correctly on pop

mpiannucci:
Brute force, but it seems to work really well. I tried some other solutions from online but this is the only one that seemed to work. I have never seen this bug before any where else.

### DIFF
--- a/NavigationReactNative/src/ios/NVSceneController.m
+++ b/NavigationReactNative/src/ios/NVSceneController.m
@@ -55,6 +55,14 @@
     }
 }
 
+- (void)viewDidAppear:(BOOL)animated
+{
+    [super viewDidAppear:animated];
+    
+    NVNavigationBarView *navigationBar = (NVNavigationBarView *) [self.view viewWithTag:NAVIGATION_BAR];
+    [navigationBar updateColors];
+}
+
 - (void)viewDidLayoutSubviews
 {
     [super viewDidLayoutSubviews];
--- a/NavigationReactNative/sample/zoom/Detail.js
+++ b/NavigationReactNative/sample/zoom/Detail.js
@@ -7,7 +7,7 @@ export default ({colors, color}) => (
   <NavigationContext.Consumer>
     {({stateNavigator}) => (
       <ScrollView contentInsetAdjustmentBehavior="automatic">
-          <NavigationBarIOS tintColor={color} titleColor={color} title="Color">
+          <NavigationBarIOS title="Color">
             <RightBarIOS>
               <BarButtonIOS systemItem="cancel" onPress={() => {
                 stateNavigator.navigateBack(1);


NavigationReactNative/src/ios/NVSceneController.m
```diff
@@ -47,14 +46,23 @@ - (void)viewWillAppear:(BOOL)animated
...
         [self.navigationItem setLargeTitleDisplayMode:navigationBar.largeTitle ? UINavigationItemLargeTitleDisplayModeAlways : UINavigationItemLargeTitleDisplayModeNever];
     }
 }
 
+- (void)viewDidAppear:(BOOL)animated {
+    [super viewDidAppear:animated];
+    
+    // This is seemingly required, otherwise the small navbar title color does not get reset when only called from viewWillAppear
+    NVNavigationBarView *navigationBar = (NVNavigationBarView *) [self.view viewWithTag:NAVIGATION_BAR];
+    [navigationBar updateTitleColors];
```
grahammendick: Do we need a separate `updateTitleColors`? Why not call `updateColors` again? I know that we don't need to update the tint colors but it doesn't cost anything

NavigationReactNative/src/ios/NVSceneController.m
```diff
@@ -47,14 +46,23 @@ - (void)viewWillAppear:(BOOL)animated
...
     if (@available(iOS 11.0, *)) {
         self.navigationController.navigationBar.prefersLargeTitles = true;
         [self.navigationItem setLargeTitleDisplayMode:navigationBar.largeTitle ? UINavigationItemLargeTitleDisplayModeAlways : UINavigationItemLargeTitleDisplayModeNever];
     }
 }
 
+- (void)viewDidAppear:(BOOL)animated {
+    [super viewDidAppear:animated];
+    
+    // This is seemingly required, otherwise the small navbar title color does not get reset when only called from viewWillAppear
```
grahammendick: Thanks but I'm not a fan of code comments. This is great detail though so you can add it to the PR's description

NavigationReactNative/src/ios/NVSceneController.m
```diff
@@ -1,15 +1,14 @@
 #import "NVSceneController.h"
-#import "NVSceneView.h"
 #import "NVNavigationBarView.h"
 #import "NVSearchBarView.h"
 
 @implementation NVSceneController
 {
-    NVSceneView *_view;
+    UIView *_view;
```
grahammendick: You're working off an old version of the `NVSceneController`. Make sure you pull from master

NavigationReactNative/src/ios/NVSceneController.m
mpiannucci: Yeah it doesn't hurt anything, was just trying to cut down on duplicate calls

mpiannucci:
Synced up with master and just made the two little changes there. Hope thats better!

grahammendick:
Looks great, thanks. In case you missed it, I released v5.5.0 with your Navigation Bar and Tab Bar color changes in!

mpiannucci:
Yes I saw!!! thank you so much!!!!

PR #296, opened 2019-09-04
Added support for navgationBar titleView on iOS.

studyroz:
Customization of `navigationItem.titleView` is common in iOS app design. This PR tries to add support for it in `navigation-react-native`.

Usage example:

~~~typescript

import { NavigationBarIOS, RightBarIOS, BarTitleViewIOS, BarButtonIOS } from 'navigation-react-native';

// Somewhere in render() method.
<NavigationBarIOS>
  <BarTitleViewIOS style={styles.titleView}>
    <Text>Customized TitleView</Text>
  </BarTitleViewIOS>
  <RightBarIOS>
    <BarButtonIOS systemItem="cancel" onPress={() => {
      stateNavigator.navigateBack(1);
    }} />
  </RightBarIOS>
</NavigationBarIOS>

~~~


grahammendick:
Thanks so much for such an interesting PR. I’m looking forward to trying to it out.

NavigationReactNative/src/ios/NVBarTitleView.m
```diff
@@ -0,0 +1,40 @@
...
+        [self.bridge.uiManager setSize:self.bounds.size forView:self];
+    }
+}
+
+- (void)didMoveToWindow {
+    [super didMoveToWindow];
+    
+    UIViewController *reactViewController = self.reactViewController;
+    if (reactViewController && self.window) {
+        self.frame = reactViewController.navigationController.navigationBar.bounds;
```
grahammendick: Why not remove the frame sizing and the `setFrame` override and leave it up to the user to set the width and height via the style prop?
```jsx
<BarTitleViewIOS style={{width:200, height: 30}}>
  <Text>Customized TitleView</Text>
</BarTitleViewIOS>
```

NavigationReactNative/src/ios/NVNavigationBarView.m
```diff
@@ -12,6 +13,14 @@ - (id)init
     return self;
 }
 
+- (void)removeReactSubview:(UIView *)subview {
+    [super removeReactSubview:subview];
+    
+    if ([subview isKindOfClass:NVBarTitleView.class]) {
+        self.reactViewController.navigationItem.titleView = nil;
+    }
+}
```
grahammendick: Let’s clear the `titleView` property using the `willMoveToSuperview` method of the `NVBarTitleView`. Then the `BarTitleViewIOS` component doesn’t have to be the child of the `NavigationBarIOS` component. I did something similar in the [NVSearchBarView](https://github.com/grahammendick/navigation/blob/753ecc6ad8c3735d583ccd2baa76a12239a5e19e/NavigationReactNative/src/ios/NVSearchBarView.m#L104)

NavigationReactNative/src/ios/NVBarTitleView.m
studyroz: Oops, this is a mistake. I thought user always want a titleView taking all the empty space.

Letting user set the style is better, I'll remove this line.

NavigationReactNative/src/ios/NVNavigationBarView.m
studyroz: 👍 This makes code more clear. There's a little layout issue after this change. I'll push the change after figure it out.

studyroz: In the current approach,`BarTitleViewIOS` style affects `NavigationBarIOS` layout properties. In some cases this leads to taking some space in content view.

For example, after adding `BarTitleViewIOS` to [`sample/zoom/Detail.js`](https://github.com/grahammendick/navigation/blob/master/NavigationReactNative/sample/zoom/Detail.js):

### Screenshot:
layout-issue-screenshot

### Hierarchy layers:
layout-issue-hierarchy

Still struggling resolving this. Please comment here if you have any idea.

NavigationReactNative/src/ios/NVBarTitleView.m
grahammendick: 👍 You can safely remove the `setFrame` method and you don't need the `bridge` anymore

NavigationReactNative/src/ios/NVNavigationBarView.m
grahammendick: Does the problem go away if you set a `width`?

studyroz: No 😞, the `BarTitleViewIOS` is a child of `NavigationBarIOS`. React Native does the layout calculation work according to the virtual dom hierarchy relation. So `NavigationBarIOS`'s `width` & `height` is affected.

NavigationReactNative/src/ios/NVBarTitleView.m
studyroz: When the `frame` of `titleView` is not set explicitly, iOS system will stretch the `titleView` to take all the space excluding `backButton` and `barButtonItems`.

And this happen outside of React's world. So here I think we need the `bridge` to notify React to correctly layout `BarTitleViewIOS`'s children.

In my experiment, this has effects when `BarTitleViewIOS`'s style is not set.

NavigationReactNative/src/ios/NVNavigationBarView.m
grahammendick: I tried adding the `BarTitleViewIOS` to the Details of zoom and it works well. The Hierarchy layers in your picture look correct to me. Are you saying it's wrong because you can't see the back button text?

NavigationReactNative/src/ios/NVBarTitleView.m
grahammendick: Can you show me how to get the problem with the children layout if we remove the `setFrame` and `bridge`, please?

NavigationReactNative/src/ios/NVNavigationBarView.m
studyroz: The hierarchy is correct, but the `NVNavigationBarView` is expected to be 0 height so that it's invisible. 
In my experiment, the height of `NVNavigationBarView` is affected by the `TitleView`.

grahammendick: Oh, thanks. I see what you mean! Sorry, I didn't look at your pictures clearly enough. Well spotted!

grahammendick: We can fix this by setting a height of 0 on the `NavigationBarIOS`.
```jsx
<NavigationBarIOS title="Color" style={{height: 0}}>
```
We shouldn't make users do this though. We should set this style inside the [NavigationBarIOS component](https://github.com/grahammendick/navigation/blob/master/NavigationReactNative/src/NavigationBarIOS.tsx) so that it's hidden from the user.

grahammendick: We can also fix it by setting position to 'absolute` instead of height to 0
```jsx
<NavigationBarIOS title="Color" style={{position: 'absolute'}}> 
```
Which one do you prefer?
Whichever we choose, we have to hide this away inside the `NavigationBarIOS component` so that the user isn't impacted

NavigationReactNative/src/ios/NVBarTitleView.m
studyroz: Of course, my code in `sample/zoom/Detail.js`:

~~~typescript
{ /* Truncated, same as zoom/Detail.js */ }
<ScrollView contentInsetAdjustmentBehavior="automatic">
  <NavigationBarIOS title="Color">
    <BarTitleViewIOS style={styles.titleView}>
      <Text style={{fontSize:18}}>Color</Text>
      <View style={{backgroundColor: color, width: 28, height: 28}}/>
    </BarTitleViewIOS>
    <RightBarIOS>
	   <BarButtonIOS systemItem='cancel' onPress={() => {
	     stateNavigator.navigateBack(1);
	   }} />
	 </RightBarIOS>
  </NavigationBarIOS>
	{ /* Truncated, same as zoom/Detail.js */ }
</ScrollView>

const styles = Style.create({
  titleView: {
    flexDirection: 'row',
    justifyContent: 'center',
    alignItems: 'center'
  },
  // Truncated, same as zoom/Detail.js
})
~~~

Here I'm adding two views as children of `BarTitleViewIOS`, they are expected to be centered in their parent.

If we keep the `setFrame` and `bridge`:
with-bridge


After removed `setFrame` and `bridge`, the `navigationBar` looks like this:
without-bridge

I add the logging code in `setFrame` and `setBounds` method and run the sample, here's the result when navigated to `Detail`:

~~~
NVBarTitleView's frame: {{0, 0}, {0, 0}}
NVBarTitleView's bounds: {{0, 0}, {375, 28}} // React layout happens here.
NVBarTitleView's frame: {{0, 0}, {279.33333333333331, 28}} // frame set by iOS
~~~

grahammendick: Thanks for such a detail reply! I'll work through your example.
By the way, I love what you're doing with the zoom example. It's a great way to demo the feature

grahammendick: The funny thing is that setting position to 'absolute' on the `NavigationBarIOS` component also fixes this problem! Try removing the `setFrame` with the following style
````jsx
<NavigationBarIOS title="Color" style={{position: 'absolute'}}>
```

NavigationReactNative/src/ios/NVNavigationBarView.m
studyroz: Cool ! I prefer `position: 'absolute'` one, let me try it.

NavigationReactNative/src/ios/NVBarTitleView.m
studyroz: Oh, that's reasonable. 
The size of `BarTitleViewIOS` is determined by it's children after it's parent's `position` set to `absolute`

I'll push the latest change. Thanks !

NavigationReactNative/src/ios/NVNavigationBarView.m
grahammendick: Just realised, we can set `position: absolute` on `BarTitleViewIOS` instead! Even better because we don't have to change `NavigationBarIOS`, right?

studyroz: Yeah, it's a better choice.

NavigationReactNative/src/BarTitleViewIOS.tsx
```diff
@@ -0,0 +1,26 @@
...
+    return (
+      <NVBarTitleView 
+        {...this.props}
+        style={[styles.barTitleView, style]}>
+        {this.props.children}
+      </NVBarTitleView>
+    )
+  }
+}
+
```
grahammendick: We can turn this into a function Component if you prefer?
```jsx
var BarTitleView = ({style, ...props}) => (
  <NVBarTitleView {...this.props} style={[styles.barTitleView, style]} />
);
```

studyroz: Yeah this makes code simpler.

studyroz:
> I think that `TitleView` is a better name than `BarTitleView`. 

Yeah I name with `Bar` prefixed to fit your naming style. `TitleView` is good.

> Why don’t we include your change to the zoom sample as part of this PR? 

I'm glad to include it. The zoom `package.json` refers dependencies to the remote one, so people clone the code before you publish a new version may get  a runtime error.

studyroz:
> `TitleView` is good.

But I prefer keep native class named to `NVBarTitleView`. It's too generic named as `NVTitleView`. the number of native classes might keep growing. There might be other type of '`TitleView`'. What's your opinion about this ?

grahammendick:
What do you think about `TitleBarIOS` and `NVTitleBarView`? 

Good point about the `package.json` pointing at the remote packages. When your PR is merged I'll get on with doing a release so the sample shouldn't be broken for long.

NavigationReactNative/src/ios/NVBarTitleView.m
```diff
@@ -0,0 +1,24 @@
...
+
+#import <React/UIView+React.h>
+
+@implementation NVBarTitleView
+
+- (void)didMoveToWindow {
+    [super didMoveToWindow];
+    
+    UIViewController *reactViewController = self.reactViewController;
+    if (reactViewController && self.window) {
```
grahammendick: Don't think we need to check that they're not nil. The [similar code in NVSearchBarView](https://github.com/grahammendick/navigation/blob/cc2e1b2be84c39d3ce88843db85edea8d48929ca/NavigationReactNative/src/ios/NVSearchBarView.m#L100) doesn't check

studyroz: OK, this check is redundant. I'll remove it.

NavigationReactNative/src/NavigationReactNative.ts
```diff
@@ -8,5 +8,6 @@ import SearchBarIOS from './SearchBarIOS';
 import TabBarIOS from './TabBarIOS';
 import TabBarItemIOS from './TabBarItemIOS';
 import SharedElementAndroid from './SharedElementAndroid';
+import TitleBarIOS from './TabBarIOS';
```
grahammendick: Ooops! This should say
```js
import TitleBarIOS from './TitleBarIOS';
```
You can test it out by running `npm run package` from the top level navigation folder and then copying the `/build/npm/navigation-react-native` folder into the `node_modules` of the zoom sample

NavigationReactNative/src/TitleBarIOS.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { requireNativeComponent, Platform, StyleSheet } from 'react-native';
+
+const NVTitleBarView = requireNativeComponent<any>('NVTitleBarView', null)
```
grahammendick: Thanks for doing the rename but this should be called `NVTitleBar` instead of `NVTitleBarView`

NavigationReactNative/src/NavigationReactNative.ts
studyroz: OK, let me test it.

build/npm/navigation-react-native/navigation.react.native.d.ts
```diff
@@ -78,6 +78,11 @@ export class LeftBarIOS extends Component { }
  */
 export class RightBarIOS extends Component { }
 
+/**
+ * Renders titleView in the UI navigation bar
+ */
+export class TitleBarIOS extends Component {  }
```
grahammendick: The `TitleBarIOS` definitions need a [style prop](https://github.com/grahammendick/navigation/blob/753ecc6ad8c3735d583ccd2baa76a12239a5e19e/NavigationReactNative/test/node_modules/%40types/navigation-react-native.d.ts#L165).

You can test that your typescript definitions work by adding your component to [navigation-react-native-test.tsx](https://github.com/grahammendick/navigation/blob/master/NavigationReactNative/test/navigation-react-native-tests.tsx). Think of the file as a unit test for the typings. Then `cd` into the 'test' folder and run `tsc --jsx react --target es6 navigation-react-native-tests.tsx`. Check the js file is generated without any errors or warnings

studyroz: Oh, didn't notice this test file.

grahammendick: You don't have to commit it, but you can use it to test your change.

grahammendick:
All that’s left is to rename `NVTitleBarViewManager` to `NVTitleBarManager` (.m and .h files). Once you do that I'll merge in this excellent PR.

I’ve really enjoyed working with you. I’m grateful for your detailed replies to my annoying questions. Please let me know if you have any suggestions on how I can improve the contribution process.

studyroz:
I appreciate your creation of this awesome project. I think your PR feedback are signs of focus to code quality, instead of annoying. I've also learnt a lot in the discussion.

If we have some basic contribution guide to this project would be better!

grahammendick:
Thanks for the nice feedback. A contribution guide is a great idea. I’m gradually getting a better feel of what needs to go in it by working with helpful contributors like you.

grahammendick:
Your hard work is now live! I [thanked you in the release notes](https://github.com/grahammendick/navigation/releases/tag/v5.7.0-NavigationReactNative)

I've updated the zoom sample to the latest `navigation-react-native`

Issue #309, opened 2019-10-12
Reg ToolbarAndroid


arunmenon1975:
React's lean core effort is going full steam and now [`ToolbarAndroid` is no longer available](https://github.com/facebook/react-native/issues/26591#issue-498739829) in the codebase though, as of today, the documentation does exist on the main doc site.

Examples in this repo that use `ToolbarAndroid`, like the Twitter sample, currently call it from `react-native` and this is bound to break in versions of `react-native` higher that `v61`, as i found out as well in my project. 

Please let me know if you are considering shipping an Android variant sometime in the future.

Pitches for having it in the library:
- `Navigation React Native` does have an iOS variant (in `TitleBarIOS` I think) so it can balance things out
- it is a key and recognizable UI artefact that perhaps is firmly tied in in premises of router solutions
- my project uses it :). Actually i suspect a rather high percentage of projects expect to use some form of a toolbar since it is a natively available across platforms and predictably familiar to users
 
To get a feel of Android native module integration with the router, my initial attempts at a crude/copy-paste integration results in errors currently. I get the dreaded 'Invariant Violation' error ("expected a string for built-in components...you likely forgot to export your component from the file it is defined in..."). I think i did the right things, at least after cross-checking with how the other modules - like the tabs -  were integrated, but somehow i keep getting the error i mentioned above. 

Steps i undertook:

- i pulled the relevant java code/source from an earlier `react-native` release, dumped them into a `toolbar` directory in `java/com/navigation/reactnative`. The files created are `DrawableWithIntrinsicSize.java`, `ReactToolbar.java` and `ReactToolbarManager.java` in the root `toolbar` directory and in a nested `events` directory i have `ToolbarClickEvent.java`. The code is verbatim from an earlier react-native version that did have the implementation. 
- updated the `NavigationPackage.java` file with the import: `import com.navigation.reactnative.toolbar.ReactToolbarManager;` and subsequent call to `new ReactToolbarManager()`. 
- the build works fine and i copy the new `navigation-react-native` directory to my `node_modules`
- despite not being type-defined as yet i went ahead with a named import of  `ToolbarAndroid` from `navigation-react-native` and attempted calling it in one of my components with a simple `<ToolbarAndroid title="Home"  />`

Im still not to sure if just pulling in the code as-is from an earlier version somehow messes up the licensing bit even if the comments and attributions are intact. My guess is it should be ok for FOSS projects.

[...]

PR #311, opened 2019-10-18
Added support for NavigationBar on Android.

emerson233:
According to the discussions [here](https://github.com/grahammendick/navigation/issues/309), I make this PR.
Check it out please.


grahammendick:
Thanks so much for your hard work. There’s lots of interesting stuff in your PR!

You created Java classes for the `Left/RightBar` the `BarButton` components. That was my first thought too, but I don’t think we should anymore. It makes it harder to implement the `TitleBar` component. This component provides a custom title view. You can see it in action on the [zoom sample’s details page](https://github.com/grahammendick/navigation/blob/7d78c28980614ec3197a3a3451dcc5a02e52e079/NavigationReactNative/sample/zoom/Detail.js#L11). The child components of the TitleBar should become child `Views` of the `Toolbar`. Then they will be automatically rendered in the place of the title. The `addView` and `removeViewAt` of the `NavigationBarManager` get complicated if the `Left/RightBar` are also children of the `Toolbar`. 

We can handle the Buttons all on the React side instead. Change the `Right/LeftBar` components to return their children (‘m only showing the android version, the iOS would remain unchanged)
```js
({children}) => children; 
```
And the `BarButton` component to return null
```js
() => null
```
Then we can loop through the children in the `NavigationBar` component on the React side and turn them into menu items. The following is a rough idea of how it might work. I haven’t run it so I’m sure it doesn’t work, but it gives you an idea of what I’m thinking about.
```js
var {children} = this.props;
var menuItems = React.Children.toArray(children).reduce((buttons, {props, type}) => {
  if (type === 'LeftBar' || type === 'RightBar')
    buttons.concat(
        React.Children.toArray(props.children)
          .map(child => ({
              ...child.props,
              image: Image.resolveAssetSource(child.props.image)
          })
    )
  return buttons;
}, []);

<NVNavigationBar menuItems={menuItems} … />
```
Then on the native side, in the `NavigateBarView` we can have a `setMenuOptions` method that behaves much like the [original setActions from React Native's Toolbar](https://github.com/facebook/react-native/blob/56751851dfefd4b99d7cf478e6d04d9dc1aa4d11/ReactAndroid/src/main/java/com/facebook/react/views/toolbar/ReactToolbar.java#L233).

Jianing93:
Looks great. I'm going to get this work.

emerson233:
Looks great. I'm going to get this work.

grahammendick:
Cool, I'm glad you like the idea!

After you've made those changes then the Title Bar component should work on Android. Rename it from `TitleBarIOS` to `TitleBar` and remove the ios `Platform` check. You can test that it works on the [Details page of the zoom sample](https://github.com/grahammendick/navigation/blob/7d78c28980614ec3197a3a3451dcc5a02e52e079/NavigationReactNative/sample/zoom/Detail.js#L11).

There's no need to update every sample to use the new `NavigationBar` component. It's enough that you update just the zoom sample. You see, once we merge this PR there will be a period of time where the updated sample won't work. It won't work until I release a new version of the package and then update the sample to use the latest package. I don't want all the samples to be broken at once. So please just update the zoom sample and back out the changes to the other samples.

emerson233:
If `BarButton` return null, then how could we pass `onPress` to native component and get callback from it?

grahammendick:
We can handle the onPress in the NavigationBar just like React Native's [Toolbar component handled the callback](https://github.com/facebook/react-native/blob/d2ef1145382b63d95f1f3374384e55487adee130/Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js#L168). Then we can get the `onPress` prop from the `menuItems` 
```jsx
var {children} = this.props;
var menuItems = React.Children.toArray(children).reduce
 ....
}, []);

<NVNavigationBar
  menuItems={menuItems}
  onPress={(positions) => menuItems[position].onPress()} />
```

emerson233:
Oh, that's a good solution. Thanks for helping.

grahammendick:
No worries. Thanks for asking!!

grahammendick:
A couple of things to look out for:

1. The `LeftBar` menu items should appear above the `RightBar` menu items. But the user might've configured the RightBar items first
```jsx
<NavigationBar>
  <RightBar>...
  <LeftBar>...
</NavigationBar>
```
So we should sort the children alphabetically by type before running the reduce
```js
React.Children.toArray(children)
  .sort((ia, b} => a.type < b.type ? -1 : (a.type > b.type ? 1 : 0))
  .reduce(
```
2. Users might want different menu items on iOS and Android so they could 
```jsx
<NavigationBar>
  <RightBar>
    <BarButton title="hello" />
    {Platform.OS === 'ios' && <BarButton title="world" />}
  </RightBar>
</NavigationBar>
```
In this case the number of children of the `RightBar` is 2 but the second child is null. So we should filter out these null children
```jsx
buttons.concat(
  React.Children.toArray(props.children)
    .filter(child => !!child)
    .map(child => ({
      ..child.props,
        image: Image.resolveAssetSource(child.props.image)
    })
)
```
I haven't tried out any of these scenarios so I might be wrong about them, but please check that your code works in both cases.

emerson233:
`React.Children.toArray` seems to filter out null children automatically.

grahammendick:
Thanks for making the changes!

1. You’ve left some old android code in the `BarButton` component. It should look like
```jsx
export default Platform.OS === 'ios' ? requireNativeComponent('NVBarButton', null) : () => null;
```
2. We can make the `NavigationBar` a bit neater by removing the android checks. There’s no harm calculating the `menuItems` and resolving the images on iOS. We can even send them as props because iOS will just ignore them. Here’s what I think it should look like. I’ve made quite a few formatting changes to give you an idea of my style preferences. You can decide whether or not to keep them in!
```jsx
var NavigationBar = ({hidden, logo, navIcon, overflowIcon, children}) => {
    if (Platform.OS == 'android' && hidden)
        return null
    var menuItems = (React.Children.toArray(children) as ReactElement<any>[])
        .filter(({type}) => (type === LeftBar || type === RightBar))
        .sort((a, b) => (a.type === b.type ? 0 : (a.type === RightBar ? 1 : -1)))
        .map(({props}) => (
            (React.Children.toArray(props.children) as ReactElement<any>[])
                .map(({props}) => ({
                    ...props,
                    image: Image.resolveAssetSource(props.image),
                })
        )
    ))
    menuItems = [].concat.apply([], menuItems)
    return (
        <NVNavigationBar
            menuItems={menuItems}
            logo={Image.resolveAssetSource(logo)}
            navIcon={Image.resolveAssetSource(navIcon)}
            overflowIcon = {Image.resolveAssetSource(overflowIcon)}
            style={{ height: 50 }}
            onActionSelected={({nativeEvent}) => {
                var onPress = menuItems[nativeEvent.position].onPress;
                if (onPress)
                    onPress()
        }}>
            {children}
        </NVNavigationBar>
    )
}
```
3. I like the way the React Native Toolbar used the ShowAsAction Java constants in JavaScript. We should copy it.  Take a look at [getExportedViewConstants](https://github.com/facebook/react-native/blob/56751851dfefd4b99d7cf478e6d04d9dc1aa4d11/ReactAndroid/src/main/java/com/facebook/react/views/toolbar/ReactToolbarManager.java#L147) and [UIManager.getViewManagerConfig](https://github.com/facebook/react-native/blob/d2ef1145382b63d95f1f3374384e55487adee130/Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js#L213)
4. We don’t need the `addTitleView` method on `NavigationBarView` because the only child is the `TitleView`. The `Right/LeftBar` are JavaScript components, not native ones, so they don’t get sent to the native side. If you expose the `toolbar`property to the manager then you can do this
```java
@Override
public void addView(NavigationBarView parent, View child, int index) {
    Toolbar.LayoutParams layoutParams = new Toolbar.LayoutParams(
        Toolbar.LayoutParams.MATCH_PARENT,
        Toolbar.LayoutParams.MATCH_PARENT,
        Gravity.CENTER
    );
    parent.toolbar.addView(child, layoutParams);
}
```
You also need to implement the `removeViewAt` method (I think you need to implement the `getChildAt` and `getChildCount` methods too?) so that the `TitleView` can be removed. You should test this by adding a `TitleView`and then removing it after a `setState`.

emerson233:
Code updated as you sugguested. Let me know if I missed something.

emerson233:
I think the [typescript definition file](https://github.com/grahammendick/navigation/blob/master/NavigationReactNative/src/node_modules/@types/react-native/index.d.ts) for react-native in this project should be update to latest version.

grahammendick:
Thanks, it's looking great!

If updating the react-native typescript helps you then please go ahead. If the method is still missing you can cast it to any ```(UIManager as any)``` instead.

Just a couple of tweaks before we move onto the next task!

1. The `BarButton` component doesn't look right to me. Shouldn't it just be what it was before?
```jsx
export default Platform.OS === 'ios' ? requireNativeComponent('NVBarButton', null) : () => null;
```
2. We can combine the two `maps` into one in the `NavigationBar` component
```jsx
.map(({props}) => (
    (React.Children.toArray(props.children) as ReactElement<any>[])
        .map(({props}) => ({
            ...props,
            image: Image.resolveAssetSource(props.image),
            show: UIManager.getViewManagerConfig(
                'NVNavigationBar',
              ).Constants.ShowAsAction[props.show],
        })
    )
))
```

grahammendick:
...Tweaks continued

3. You don't need the `titleViews` property on `NavigationBarView`. You can use the toolbar directly instead, for example,
```java
@Override
public int getChildCount(NavigationBarView parent) {
    return parent.toolbar.getChildCount();
}
```

emerson233:
Toolbar's children contains it's built-in menuItem and titleView. So maybe `parent.toolbar.getChildCount()`  is not suitable here?

grahammendick:
Interesting. What about `parent.toolbar.getChildCount() - 2`? Is that confusing?

emerson233:
Yeah it's a bit confusing. I don't think it's a good practice.

grahammendick:
Hold on, what's wrong with returning `parent.toolbar.getChildCount()`? If it's got two children then we should return 2 right? That should still work

emerson233:
Emm...You are right. Nothing wrong with it. I am going to fix it.

emerson233:
Note that there's a left margin on the `Toolbar` by default. So the custom `TitleBarView` is not exactly centered. Should we remove this left padding? [Related question on StackOverflow](https://stackoverflow.com/questions/27354812/android-remove-left-margin-from-actionbars-custom-layout)`

grahammendick:
Is it just that it's the wrong color? Is it the color of the `AppBar` showing through? Our next task is to work on the color props so maybe we should leave it until after that?

emerson233:
It'not about wrong color. Because of this left padding,  something like `justifyContent: 'center'` or  `alignItems: 'center'` on react side does not align items quite right.

grahammendick:
Cool, let's remove it then

grahammendick:
Thanks again for all your hard work. Just the `BarButton.tsx` change to make. I think you can delete your changes because shouldn't it just be this?
```jsx
export default Platform.OS === 'ios' ? requireNativeComponent('NVBarButton', null) : () => null;
```

emerson233:
How do you read image from the image path on the iOS side? Shouldn't we call `Image.resolveAssetSource` before we pass it to native component? Or it's not related to this PR, so we shouldn't change it now.

emerson233:
After removing the left padding, the built-in title on `Toolbar` looks not so good.
image
- Should we remove the left padding only if there is a cutome title view?
- How to know there is a custom title view?
- How to restore the previous padding properly?

grahammendick:
> How do you read image from the image path on the iOS side? Shouldn't we call Image.resolveAssetSource before we pass it to native component? 

iOS doesn't need `resolveAssetSource`. On iOS, you declare the [property as a UIImage](https://github.com/grahammendick/navigation/blob/7d78c28980614ec3197a3a3451dcc5a02e52e079/NavigationReactNative/src/ios/NVBarButtonView.m#L24) and React Native takes care of it all.

grahammendick:
How are you removing the padding? Is it like setting the `contentInsetStart` prop to 0 on the React Native Toolbar?

emerson233:
Calling `toolbar.setContentInsetsAbsolute(0, 0)` on Android side.

grahammendick:
Take a look at how the [React Native toolbar gets the default content insets](https://github.com/facebook/react-native/blob/56751851dfefd4b99d7cf478e6d04d9dc1aa4d11/ReactAndroid/src/main/java/com/facebook/react/views/toolbar/ReactToolbarManager.java#L182). You'll need to reset them if the user removes the `TitleBarView`. 

We can probably remove the insets in `addView` and restore them in `removeViewAt`. You'll need to check it works if the user changes the `key` prop of the `TitleBar` after a setState because this will call both an add **and** a remove on the native side!
```jsx
<NavigaitonBar> 
  <TitleBar key="a">...
</NavigationBar>
```
If you prefer we can do it on the JavaScript side. We can check the `React.Children` for the presence of a `TitleBar` and then send down a `hasTitleBar` prop?

emerson233:
It works correctly when I change the `key` prop.
I prefer to do it on the Android side. Please check these [codes](https://github.com/grahammendick/navigation/pull/311/commits/1943ed5a579cfe7694e7812c3c9f699aacdad6eb).

grahammendick:
Looks great, I prefer to do it on the native side, too! Just wondering if there's an easier way to get the inset defaults. When we create the Toolbar can we get the insets and cache them in the View? Are they applied after the Toolbar is created or are they applied later?

grahammendick:
Something like this
```java
public NavigationBarView(Context context) {
    super(context);
    toolbar = new Toolbar(context);
    contentInsetStart = toolbar.getContentInsetStart();
```

emerson233:
Fine. Let's do it.
I am planning to add `barTintColor`, `tintColor`, `titleColor` props to Android side too. Is there other props needed to add?

grahammendick:
Cool, thanks. Sorry I didn't think of it earlier.

I think the color props are the only ones we're missing. Can we chat about how to implement them first. I'm keen to hear your ideas on how we should go about it.

emerson233:
```java
@ReactProp(name = "barTintColor", customType = "Color")
public void setBarTintColor(NavigationBarView view, @Nullable Integer barTintColor) {
    view.setBackgroundColor(barTintColor != null ? barTintColor : defaultColor);
}

@ReactProp(name = "titleColor", customType = "Color")
public void setTitleColor(NavigationBarView view, @Nullable Integer textColor) {
    view.toolbar.setTitleTextColor(textColor != null ? textColor : defaultColor);
}
```
Something like these.
I don't see how to implement `tintColor` on the Android side.

grahammendick:
Can we use the style prop to set the background color instead?
 ```jsx
style={{backgroundColor: barTintColor}}
```
I like ```view.toolbar.setTitleTextColor```. I think getting the default title color could be tricky ;)!

I don’t think we should provide an android implementation of the `tintColor` prop. This prop changes the color of the buttons on iOS. We could change the color of the menu items but it’s a lot of work and I’m not sure it’s worth it. The problem is that I can’t find a way to programmatically change the background color of the menu. So I don’t see the point of changing the menu item colors if we can’t change the menu background color at the same time! Seems to make more sense to change them both through style.xml. What do you think?

**Comment by grahammendick** [comment]
I guess the problem with ```style={{backgroundColor: barTintColor}}``` is that it won't reset to the default toolbar background color? These default resets are confusing me!!

**Comment by emerson233** [comment]
Agree. Will post possible solutions to discuss here after some coding.

**Comment by emerson233** [comment]
See [How ReactToolbarManager gets default text color](https://github.com/facebook/react-native/blob/56751851dfefd4b99d7cf478e6d04d9dc1aa4d11/ReactAndroid/src/main/java/com/facebook/react/views/toolbar/ReactToolbarManager.java#L212)
Copy and paste `getDefaultColors` implementation from there, and codes will look like
```java
@ReactProp(name = "barTintColor", customType = "Color")
public void setBarTintColor(NavigationBarView view, @Nullable Integer barTintColor) {
    if (barTintColor != null) {
        view.toolbar.setBackgroundColor(barTintColor);
    } else {
        view.toolbar.setBackground(null);
    }
}

@ReactProp(name = "titleColor", customType = "Color")
public void setTitleColor(NavigationBarView view, @Nullable Integer textColor) {
    int[] defaultColors = getDefaultColors(view.getContext());
    view.toolbar.setTitleTextColor(textColor != null ? textColor : defaultColors[0]);
}
```
Seems good enough. What do you think?

grahammendick:
Can we reformat the default title text color code so that it's consistent with [similar code in the SceneNavigator](https://github.com/grahammendick/navigation/blob/7d78c28980614ec3197a3a3451dcc5a02e52e079/NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/SceneNavigator.java#L31), please? Also, perhaps we should move it into the `NavigationBarView` constructor so that it's consistent with the content insets
```java
public NavigationBarView(Context context) {
    super(context);
    toolbar = new Toolbar(context);
    contentInsetStart = toolbar.getContentInsetStart();
   /* Get default title text color */
    titleTextColor = titleTextAppearance.getColor(0, Color.BLACK);
```
Is there any difference between your `setBarTintColor` method and 
```jsx
style={{backgroundColor: barTintColor}}
```
Does setting `barTIntColor` to null give different results?

emerson233:
Setting `backgroudColor` on the React side:
image
image
Using `view.toolbar.setBackgroundColor`:
image
image
I am changing color from "red" to `undefined` here by clicking right button.

grahammendick:
Cool, thanks for showing me the difference in such a clear way!
I guess the difference is that, in the first case, the `backgroundColor` is applied to the `AppBarLayout` and not  the `Toolbar`? Anyway, I'm happy with your suggestion for the `setBarTintColor` method.

emerson233:
As for `titleTextAppearance`, is it possible to get its value from `Toolbar`. I didn't see a getter for it in `Toolbar`.

grahammendick:
I couldn't find one either :(

grahammendick:
I mean we should format the `getDefaultColors` code so it looks like the [similar code from the Navigation router codebadse](https://github.com/grahammendick/navigation/blob/7d78c28980614ec3197a3a3451dcc5a02e52e079/NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/SceneNavigator.java#L31). Just for consistency - even though we're taking code from React Native's Toolbar we should format it so it's consistent with code from the Navigation router.

emerson233:
You mean using `android.R.attr.toolbarStyle` instead of `getIdentifier(context, "toolbarStyle")` and holding a `titleTextAppearance` reference?

grahammendick:
Nope, The code is fine, I just don't like the formatting. I don't like all the whitespace and carriage returns. Sorry for the confusion.

emerson233:
So now codes look like
```java
// NavigationBarView.java
public NavigationBarView(Context context) {
    // ...
    defaultContentInsetStart = toolbar.getContentInsetStart();
    defaultContentInsetEnd = toolbar.getContentInsetEnd();
    defaultTitleTextColor = getDefaultTitleTextColor(context);
}

private static int getDefaultTitleTextColor(Context context) {
    Resources.Theme theme = context.getTheme();
    TypedArray toolbarStyle = null;
    TypedArray textAppearances = null;
    TypedArray titleTextAppearance = null;

    try {
        toolbarStyle = theme.obtainStyledAttributes(new int[] {getIdentifier(context, "toolbarStyle")});
        int toolbarStyleResId = toolbarStyle.getResourceId(0, 0);
        textAppearances = theme.obtainStyledAttributes(toolbarStyleResId, new int[] {
                getIdentifier(context, "titleTextAppearance")
        });

        int titleTextAppearanceResId = textAppearances.getResourceId(0, 0);
        titleTextAppearance = theme.obtainStyledAttributes(titleTextAppearanceResId, new int[]{android.R.attr.textColor});
        int titleTextColor = titleTextAppearance.getColor(0, Color.BLACK);
        return titleTextColor;
    } finally {
        recycleQuietly(toolbarStyle);
        recycleQuietly(textAppearances);
        recycleQuietly(titleTextAppearance);
    }
}

private static void recycleQuietly(@Nullable TypedArray style) {
    if (style != null) {
        style.recycle();
    }
}

private static int getIdentifier(Context context, String name) {
    return context.getResources().getIdentifier(name, "attr", context.getPackageName());
}

// NavigationBarManager.java
@Override
public void addView(NavigationBarView parent, View child, int index) {
    Toolbar.LayoutParams layoutParams = new Toolbar.LayoutParams(
            Toolbar.LayoutParams.MATCH_PARENT,
            Toolbar.LayoutParams.MATCH_PARENT
    );
    parent.toolbar.addView(child, index, layoutParams);
    parent.toolbar.setContentInsetsRelative(0, 0);
}

@Override
public void removeViewAt(NavigationBarView parent, int index) {
    parent.toolbar.removeViewAt(index);
    parent.toolbar.setContentInsetsRelative(parent.defaultContentInsetStart, parent.defaultContentInsetEnd);
}

@ReactProp(name = "barTintColor", customType = "Color")
public void setBarTintColor(NavigationBarView view, @Nullable Integer barTintColor) {
    if (barTintColor != null) {
        view.toolbar.setBackgroundColor(barTintColor);
    } else {
        view.toolbar.setBackground(null);
    }
}

@ReactProp(name = "titleColor", customType = "Color")
public void setTitleColor(NavigationBarView view, @Nullable Integer textColor) {
    view.toolbar.setTitleTextColor(textColor != null ? textColor : view.defaultTitleTextColor);
}

```

grahammendick:
Thanks, looking good, but I'd go a bit further still
```java
private static int getDefaultTitleTextColor(Context context) {
  Resources.Theme theme = context.getTheme();
  TypedArray toolbarStyle = theme.obtainStyledAttributes(new int[] {getIdentifier(context, "toolbarStyle")});
  int toolbarStyleResId = toolbarStyle.getResourceId(0, 0);
  toolbarStyle.recycle();
  TypedArray textAppearances = theme.obtainStyledAttributes(toolbarStyleResId, new int[] {getIdentifier(context, "titleTextAppearance")});
  int titleTextAppearanceResId = textAppearances.getResourceId(0, 0);
  textAppearances.recycle();
  TypedArray titleTextAppearance = theme.obtainStyledAttributes(titleTextAppearanceResId, new int[]{android.R.attr.textColor});
  int titleTextColor = titleTextAppearance.getColor(0, Color.BLACK);
  titleTextAppearance.recycle();
  return titleTextColor;
}
```

grahammendick:
Cool. Do you want to hear what's next?

emerson233:
Yeah, go ahead.

grahammendick:
We should look at the `navIcon` prop next.  Let’s rename it to `backImage` and have an associated `onBackPress` prop. We can have a default implementation that navigates back one scene. 
```jsx
var NavigationBar = ({onBackPress, stateNavigator}) => {
    
    <NVNavigationBar
        onBackPress={() => {
            if (onBackPress)
                onBackPress();
            if (!onBackPress && stateNavigator.canNavigateBack(1))
                stateNavigator.navigateBack(1)
        }}
    >
        {children}
    </NVNavigationBar>
```
The `stateNavigator` is available from the React context.
```jsx
export default props => (
    <NavigationContext.Consumer>
        {({stateNavigator}) => <NavigationBar stateNavigator={stateNavigator} {...props} />}
    </NavigationContext.Consumer>
)
```
What do you think?

emerson233:
I agree on adding a `onPress` event on `navIcon`, but I don't think it's good idea to treat it as backImage by default. It could be a menu icon or something else.
I think we could offer a `backIcon` prop and if it has value, pass it to `navIcon`. Furthermore, we could also bring this prop to the iOS side in the future.

grahammendick:
That's an interesting idea about using it on iOS in the future!

OK, we won't rename it to `backImage`. I think we should rename it to `image` or  `navigationImage` though, for consistency with `BarButton` image prop. Then what should we call the handler prop? `onNavigationPress` or `onPress`.or something else?

emerson233:
I will go with `onNavigationPress `.
```javascript
<NVNavigationBar
    navigationImage={Image.resolveAssetSource(navigationImage)}
    onNavigationPress={onNavigationPress}
    {...otherProps}
}}>
    {children}
</NVNavigationBar>
```

grahammendick:
Cool

grahammendick:
What do you think about doing a default back navigation if `onNavigationPress` is null?

emerson233:
If you insist to set a default back navigation on the `navigationImage`, I recommend to expose a prop named `backImage` to user and pass it to `NVNavigationBar`'s `navIcon`.
```javascript
var NavigationBar = ({backImage, onBackPress, stateNavigator}) => {
    <NVNavigationBar
        navIcon={Image.resolveAssetSource(backImage)}
        onBackPress={() => {
            if (onBackPress)
                onBackPress();
            if (!onBackPress && stateNavigator.canNavigateBack(1))
                stateNavigator.navigateBack(1)
        }}
        {...otherProps}
    }}>
        {children}
    </NVNavigationBar>
}
```

grahammendick:
Nope, let's leave it. There's no need for a default back 👍

emerson233:
Take a look at the latest changes. I added the `onPress` event before and I just forgot. Now I rename it as discussed.

grahammendick:
It's looking great, thanks. 

Next, can we work on backward compatibility. I want to bring in the `NavigationBar` in a minor release. That means that we mustn't break existing code. So we need to keep `NavigationBarIOS`,  for example. You can copy [the approach I took when I made TabBarIOS cross platform](https://github.com/grahammendick/navigation/blob/7d78c28980614ec3197a3a3451dcc5a02e52e079/NavigationReactNative/src/NavigationReactNative.ts#L15). 

Also, we should only update one sample. The zoom sample is the best because it has `BarButtons`. Can you undo the changes to other samples because I don't want them to break when we merge this PR. 

Any questions then let me know.

emerson233:
Sure. Going to push a commit later.

emerson233:
Now the twitter sample is breaking because `ToolbarAndroid` is removed from react-native. Should we modify this one too?

grahammendick:
The Twitter sample shouldn't break because [it's on react native v60.5](https://github.com/grahammendick/navigation/blob/ed629cabebe0e417d886cf7b6059366f8c63000e/NavigationReactNative/sample/twitter/package.json#L15).

emerson233:
Alright. Sample updated.

grahammendick:
Thanks, it feels like we're quite close now!

Can we go through the Java and bring it in line with the Navigation router style/format. For example,  can we 
1. Remove the 'm' prefix from private fields, e..g, `mWidth` in `IconImageInfo`
2. Remove all comments
3. Remove blank lines inside methods

emerson233:
Menu items on Android require title field to be non-null. Should we provide a default empty string if it's null? 
```java
String title = menuItemProps.hasKey(PROP_ACTION_ICON)
        ? menuItemProps.getString(PROP_ACTION_TITLE)
        : "";
```
Or we should declare this prop as required on the `.d.ts` file?
```
title?: string; -> title: string;
```

grahammendick:
A default empty string is better. It can't be a mandatory prop because the user might only be using the NavigationBar on iOS

grahammendick:
I'm going out in a bit so I'm posting the next task while I have the chance. 

For backward compatibility, we need to add the IOS components (`NavigaitonBarIOS` etc.) back into the Typescript definitions. Let’s follow the approach I took with `TabBarIOS`, where I point them at the new props (so we don’t need to add back `NavigationBarIOSProps`)
```ts
export class NavigaitonBarIOS extends Component<NavigaitonBarProps> {}
```
Can we also tweak the comments for the new props
```ts
/**
 * The logo
 */
logo?: ImageURISource;
/**
 * The menu overflow image
 */
overflowImage?: ImageURISource;
/**
 * The navigation button image
 */
navigationImage?: ImageURISource;
/**
 * Handles navigation button click events
 */
onNavigationPress?: () => void
```

Also, we’re missing the `show` prop on the `BarButton` component. 
```ts
/**
 * Determines when this item should appear in the navigation bar
 */
show?: 'ifroom' | 'never' | 'always';
```

Also, thanks for changing the Typescript tests in `navigation-react-native-tests.tsx`. Even though they're just checking the TypeScript definitions I like to keep them as real as possible. So we should move the `NavigationBar` to the top of each scene just like you did in the zoom sample.

emerson233:
Updated!

grahammendick:
Thanks! 
Will you go through the Java and bring it in line with the Navigation router style/format. For example,  let's 
1. Remove the 'm' prefix from private fields, e..g, `mWidth` in `IconImageInfo`
2. Remove all comments
3. Remove blank lines inside methods

emerson233:
It has been done in [this commit](https://github.com/grahammendick/navigation/pull/311/commits/9e1926c79ef5bc323b6bd772bf8bed99d79c8c18).

grahammendick:
My bad! Thanks!

grahammendick:
Have you tried implementing a transparent Navigation Bar with the content scrolling underneath it? What are the other common Android Navigation Bar patterns that we should test out?

emerson233:
Tested the transparent Navigation Bar just now. We can adjust style like `marginTop` of `ScrollView` in React to draw it under `Toolbar`. We have to set `AppBarLayout`'s backgroundColor to transparent.
```java
@ReactProp(name = "barTintColor", customType = "Color")
public void setBarTintColor(NavigationBarView view, @Nullable Integer barTintColor) {
    if (barTintColor != null) {
        // view.toolbar.setBackgroundColor(barTintColor);
        view.setBackgroundColor(barTintColor);
    } else {
        // view.toolbar.setBackground(null);
        view.setBackgroundColor(null);
    }
}
```
image
To remove shadow (https://stackoverflow.com/questions/50277982/failed-to-make-appbarlayout-transparent), it looks like
image

emerson233:
Also, [collapsing-toolbar-layout](https://material.io/develop/android/components/collapsing-toolbar-layout/) is an interesting pattern to implement.

grahammendick:
What about setting the background color on the AppBarLayout and the Toolbar at the same time? Does it make a difference? 
```java
@ReactProp(name = "barTintColor", customType = "Color")
public void setBarTintColor(NavigationBarView view, @Nullable Integer barTintColor) {
    if (barTintColor != null) {
        view.toolbar.setBackgroundColor(barTintColor);
        view.setBackgroundColor(barTintColor);
    } else {
        view.toolbar.setBackground(null);
        view.setBackgroundColor(null);
    }
}
```
The shadow in your picture is in an odd place, why is there a gap between the shadow and the toolbar?
Here's another [way to remove the shadow](https://stackoverflow.com/questions/43255496/remove-android-widget-toolbar-shadow). Should we add a 'shadow' boolean prop to the NavigationBar and default to true?

Let me know how you get on with the collapsing toolbar.

emerson233:
https://stackoverflow.com/a/37992366/3480295
`appBayLayout.setElevation(0)` does not remove shadow programmatically. It does work when setting the elevation in the XML though.
And Setting backgroundColor for both views does not make a difference.

emerson233:
As for collapsing toolbar, I haven’t thought this through yet. Just an idea.

grahammendick:
We don't need to work on collapsing toolbar for this release. What do you think?

grahammendick:
I think we should have a transparent toolbar for this release, though.
If we're going to set the backgroundColor of the AppBarLayout instead of the Toolbar then can we use the style prop instead?
```jsx
style={{backgroundColor: 'blue'}}
```
Instead of adjusting the `marginTop` would a transparent navigation bar work better if we set the position of the navigation bar to 'absolute'?

grahammendick:
Whenever we set a transparent `barTintColor` then the `AppBarLayout` shadow shows through. So if the `barTintColor` is semi-transparent (alpha < 255) then we should hide the shadow. That way we can support a transparent navigation bar without adding a 'shadow' prop.
```java
@ReactProp(name = "barTintColor", customType = "Color")
public void setBarTintColor(NavigationBarView view, @Nullable Integer barTintColor) {
    if (barTintColor != null) {
        view.setBackgroundColor(barTintColor);
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP && Color.alpha(barTintColor) < 255)
            view.setOutlineProvider(null);
    } else {
        view.setBackground(null);
    }
}
```

emerson233:
> We don't need to work on collapsing toolbar for this release. What do you think?

Yeah. We could implement it in future releases.



> Whenever we set a transparent `barTintColor` then the `AppBarLayout` shadow shows through. So if the `barTintColor` is semi-transparent (alpha < 255) then we should hide the shadow. That way we can support a transparent navigation bar without adding a 'shadow' prop. We have to revert to the default outline provider if it's not transparent (cache the default outline provider in the NavigationView constructor like we do with the other defaults). What do you think?
> 
> ```java
> @ReactProp(name = "barTintColor", customType = "Color")
> public void setBarTintColor(NavigationBarView view, @Nullable Integer barTintColor) {
>     view.setBackground(null);
>     view.setOutlineProvider(view.defaultOutlineProvider);
>     if (barTintColor != null) {
>         view.setBackgroundColor(barTintColor);
>         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP && Color.alpha(barTintColor) < 255)
>             view.setOutlineProvider(null);
>     }
> }
> ```

Looks great. Going to test it out to see if it works.

emerson233:
> Setting `backgroudColor` on the React side:
> image
> image
> Using `view.toolbar.setBackgroundColor`:
> image
> image
> I am changing color from "red" to `undefined` here by clicking right button.

Must set `background` of `AppBarLayout` back to default to resolve this problem when setting state to clear `barTintColor`.

NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/IconResolver.java
```diff
@@ -0,0 +1,143 @@
...
+
+import com.facebook.drawee.backends.pipeline.Fresco;
+import com.facebook.drawee.controller.BaseControllerListener;
+import com.facebook.drawee.interfaces.DraweeController;
+import com.facebook.drawee.view.DraweeHolder;
+import com.facebook.imagepipeline.image.ImageInfo;
+import com.facebook.imagepipeline.image.QualityInfo;
+import com.facebook.react.bridge.ReadableMap;
+import com.facebook.react.uimanager.PixelUtil;
+import com.facebook.react.views.toolbar.DrawableWithIntrinsicSize;
```
grahammendick: We need to copy over DrawableWithIntrinsicSize because [it's been removed from React Native 0.61](https://github.com/facebook/react-native/commit/cbbbb455dda313cca1a1763cc5453079ce87bc73)

We need to upgrade the zoom sample to the latest React Native to make sure that it works. I upgrade by deleting the zoom sample and recreating it but you need to be careful if you want to do it this way because you might lose the deep linking functionality. If you want to delete and recreate then let me know and I'll write out the steps required.

NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/NavigationBarView.java
```diff
@@ -0,0 +1,240 @@
...
+            defaultOutlineProvider = getOutlineProvider();
+        }
+        defaultBackground = getBackground();
+
+        iconResolver = new IconResolver(context);
+        mLogoControllerListener = iconResolver.new IconControllerListener(mLogoHolder) {
+            @Override
+            protected void setDrawable(Drawable d) {
+                toolbar.setLogo(d);
+                requestLayout();
```
grahammendick: The calls to `requestLayout` don't match with the original React Native Toolbar code. The setDrawable of the logp, navigation and overflow icons [don't call requestLayout in the original](https://github.com/facebook/react-native/blob/56751851dfefd4b99d7cf478e6d04d9dc1aa4d11/ReactAndroid/src/main/java/com/facebook/react/views/toolbar/ReactToolbar.java#L144).  But the setDrawable of the menu item images do call [requestLayout in the original](https://github.com/facebook/react-native/blob/56751851dfefd4b99d7cf478e6d04d9dc1aa4d11/ReactAndroid/src/main/java/com/facebook/react/views/toolbar/ReactToolbar.java#L100). In our code this is reversed. The logo calls requestLayout but the menu items don't. Any reason for this change?

NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/IconResolver.java
emerson233: File added.
I am not updating the zoom sample to the latest React Native because after doing `npm install`, lots of lines in `package-lock.json` changed and I don't know if I am doing right. Maybe because we use different npm version?

emerson233: Nevertheless, I tested it out and the sample is working on React Native v0.61 now.

grahammendick: No problem, just let me know how you tested it on React Native v0.61

NavigationReactNative/src/NavigationBar.tsx
```diff
@@ -0,0 +1,43 @@
...
+            )
+        ))
+        .reduce((a, b) => a.concat(b), [])
+    return (
+        <NVNavigationBar
+            menuItems={menuItems}
+            logo={Image.resolveAssetSource(logo)}
+            navigationImage={Image.resolveAssetSource(navigationImage)}
+            overflowImage={Image.resolveAssetSource(overflowImage)}
+            style={[Platform.OS === 'android' ? { height: 50 } : undefined, style]}
```
grahammendick: Change to the style prop as follows
```jsx
style={Platform.OS === 'android' ? {height: 56} : undefined}
```
There are two changes here
1. The height is 56 to match the [material design specs](https://material.io/components/app-bars-top/#specs)
2. Removed the user style prop. If we let the user customise the style then it we could accidentally break their code when we add new features, e.g., collapsing toolbar.

NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/NavigationBarView.java
```diff
@@ -0,0 +1,252 @@
...
+    private void attachDraweeHolders() {
+        mLogoHolder.onAttach();
+        mNavIconHolder.onAttach();
+        mOverflowIconHolder.onAttach();
+        mActionsHolder.onAttach();
+    }
+
+    void setTitle(String title) {
+        toolbar.setTitle(title);
+    }
```
grahammendick: Remove this method and directly call ```view.toolbar.setTitle``` from the NavigationBarManager instead

NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/NavigationBarView.java
```diff
@@ -0,0 +1,252 @@
...
+    };
+
+    @Override
+    public void requestLayout() {
+        super.requestLayout();
+        post(mLayoutRunnable);
+    }
+
+    private static int getDefaultTitleTextColor(Context context) {
+        Resources.Theme theme = context.getTheme();
```
grahammendick: Don't make this method static and don't pass the context in
```java
private int getDefaultTitleTextColor() {
    Resources.Theme theme = getContext().getTheme();
```

build/npm/navigation-react-native/navigation.react.native.d.ts
```diff
@@ -69,12 +69,48 @@ export interface NavigationBarIOSProps {
...
+     */
+    navigationImage?: ImageURISource;
+    /**
+     * Handles navigation button click events
+     */
+    onNavigationPress?: () => void
+    /**
+     * Determines when this item should appear in the navigation bar
+     */
+    show?: 'ifRoom' | 'never' | 'always';
```
grahammendick: The `show` prop should be on the `BarButtonProps`. Don't forget to copy the changes into the other [navigation-react-native.d.ts file](https://github.com/grahammendick/navigation/blob/master/NavigationReactNative/test/node_modules/%40types/navigation-react-native.d.ts)

NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/NavigationBarManager.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @ReactProp(name = "navigationImage")
+    public void setNavIcon(NavigationBarView view, @Nullable ReadableMap navIcon) {
+        view.setNavIconSource(navIcon);
+    }
+
+    @ReactProp(name = "overflowImage")
+    public void setOverflowIcon(NavigationBarView view, @Nullable ReadableMap overflowIcon) {
+        view.setOverflowIconSource(overflowIcon);
+    }
```
grahammendick: When the `overflowImage` prop is null then this removes the default overflow icon. We need to store the `defaultOverflowIcon` in the View constructor and restored it when null

NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/NavigationBarManager.java
```diff
@@ -0,0 +1,124 @@
...
+        view.setMenuItems(menuItems);
+    }
+
+    @ReactProp(name = "barTintColor", customType = "Color")
+    public void setBarTintColor(NavigationBarView view, @Nullable Integer barTintColor) {
+        if (barTintColor != null) {
+            view.setBackground(null);
+            view.toolbar.setBackgroundColor(barTintColor);
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP && Color.alpha(barTintColor) < 255)
+                view.setOutlineProvider(null);
```
grahammendick: We need to restore the default outline provider if the color is set and not transparent. We might be changing from a transparent to a non-transparent color after a `setState`

NavigationReactNative/src/NavigationBar.tsx
```diff
@@ -0,0 +1,43 @@
...
+        return null
+    var menuItems = (React.Children.toArray(children) as ReactElement<any>[])
+        .filter(({type}) => (type === LeftBar || type === RightBar))
+        .sort((a, b) => (a.type === b.type ? 0 : (a.type === RightBar ? 1 : -1)))
+        .map(({props}) => (
+            (React.Children.toArray(props.children) as ReactElement<any>[])
+                .map(({props}) => ({
+                    ...props,
+                    show: Platform.OS === 'android'
+                     ? UIManager.getViewManagerConfig('NVNavigationBar').Constants.ShowAsAction[props.show]
```
grahammendick: Cast `UIManager` to `any` to avoid the typescript error
```tsx
(UIManager as any).getViewManagerConfig('NVNavigationBar')...
```

NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/NavigationBarView.java
```diff
@@ -0,0 +1,252 @@
...
+    Drawable defaultBackground;
+
+    private final DraweeHolder mLogoHolder = DraweeHolder.create(createDraweeHierarchy(), getContext());
+    private final DraweeHolder mNavIconHolder = DraweeHolder.create(createDraweeHierarchy(), getContext());
+    private final DraweeHolder mOverflowIconHolder = DraweeHolder.create(createDraweeHierarchy(), getContext());
+
+    private IconResolver.IconControllerListener mLogoControllerListener;
+    private IconResolver.IconControllerListener mNavIconControllerListener;
+    private IconResolver.IconControllerListener mOverflowIconControllerListener;
+    private final MultiDraweeHolder<GenericDraweeHierarchy> mActionsHolder =
```
grahammendick: Remove the 'm' prefix for these variables

NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/DrawableWithIntrinsicSize.java
```diff
@@ -0,0 +1,26 @@
+package com.navigation.reactnative;
+
+import android.graphics.drawable.Drawable;
+
+import com.facebook.drawee.drawable.ForwardingDrawable;
+import com.facebook.imagepipeline.image.ImageInfo;
+
+public class DrawableWithIntrinsicSize extends ForwardingDrawable implements Drawable.Callback {
+    private final ImageInfo mImageInfo;
```
grahammendick: Remove the 'm' prefix for this variable

NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/IconResolver.java
grahammendick: I've created a [PR that updates the zoom sample to v0.61](https://github.com/emerson233/navigation/pull/1). Please merge it into this PR if you're happy

emerson233: > No problem, just let me know how you tested it on React Native v0.61

Just changed `package.json` like this.
```
"react": "16.9.0",
"react-native": "0.61.0"
```
And doing a `npm install`.

emerson233:
All issues resolved. Let me know if I miss something.

build/npm/navigation-react-native/navigation.react.native.d.ts
grahammendick: The `show` prop should be on the `BarButtonProps`not the `NavigationBarProps`

NavigationReactNative/src/NavigationBar.tsx
grahammendick: If the user provides a style prop then at the moment we're still rendering it because we include the ```{...otherProps}```. We can either:
1. Move {...otherProps} before the our style prop so that the user style is ignored
2. Destructure the user style prop and ingore it

NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/TitleBarViewManager.java
```diff
@@ -0,0 +1,20 @@
+package com.navigation.reactnative;
+
+import com.facebook.react.uimanager.ThemedReactContext;
+import com.facebook.react.uimanager.ViewGroupManager;
+
+import javax.annotation.Nonnull;
+
+public class TitleBarViewManager extends ViewGroupManager<TitleBarView> {
```
grahammendick: Rename from `TitleBarViewManager` to `TitleBarManager`

grahammendick:
Hey @emerson233 we're getting close now! I've prepared the [Navigation Bar docs](https://github.com/grahammendick/navigation/pull/317) for the next release. Let me know if you need any help merging in [my iOS fix PR](https://github.com/emerson233/navigation/pull/2) or with the few outstanding tasks. Thanks for all your hard work!

emerson233:
No problems. Thanks for your thoughtful help.

grahammendick:
There are a couple of comments that you haven't responded to:

1. The `show` prop  should be on the `BarButtonProps` not the `NavigationBarProps` in the Typescript typings
2. The `NavigationBar` component renders any user supplied styles ```<NavigationBar style={{..}}/>```. It shouldn't do that. The easiest way to prevent this is to revert [the commit that removed the style destructure](https://github.com/grahammendick/navigation/pull/311/commits/120b29196f9ae0b97b26cc575176e5b233aa8da2)

emerson233:
Fixed. Have a look.

grahammendick:
Looks good, thanks!

grahammendick:
If there's a `navigationImage`and a `TitleBar` then the title view isn't centred. This is what I tried on the zoom details scene. Do you know why?
```jsx
<NavigationBar
  title="Color"
  navigationImage={{uri: 'https://facebook.github.io/react-native/img/tiny_logo.png'}}
  onNavigationPress={() => {
    stateNavigator.navigateBack(1);
  }}>
```

emerson233:
Maybe it's not recommended to use both custom titleView and navIcon on the toolbar? I don't know.

grahammendick:
@emerson233 Can you check over [my PR that removes the title centering](https://github.com/emerson233/navigation/pull/3) and merge it in if you're happy? Let me know if have any questions about it. Thanks!

emerson233:
Merged.

grahammendick:
Thanks again for such a great PR. I’m really happy to merge it in. Fixes #309.

I tried to help by submitting my own mini-PRs to merge into this one, but I think they ended up slowing things down. I’ll probably avoid doing that in future PRs. How do feel about them? Did you find them helpful or would you have preferred that I’d stuck to review comments only?

emerson233:
It's not slowing things down. Sometimes I cannot respond to your comments in time because of the time difference between us, and your PRs are really helping me out.
Anyway, thanks for all your patience and kindness.

PR #312, opened 2019-10-20
Fix a bug when another library push fragments in the middle

melkayam92:
For example, when using react-native-fast-image, it uses another fragment to load the images. so when getting the last fragment we get their instance which is another type from SceneFragment and we get an exception. 

This fix simply traverses from the end and finds the last SceneFragment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ node_modules
 .DS_Store
 .project
 org.eclipse.buildship.core.prefs
+.classpath
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@ node_modules
 !Navigation*/src/node_modules
 !Navigation*/test/node_modules
 .DS_Store
-
+.project
+org.eclipse.buildship.core.prefs
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,3 @@ node_modules
 !Navigation*/src/node_modules
 !Navigation*/test/node_modules
 .DS_Store
-.project
-org.eclipse.buildship.core.prefs
-.classpath
--- a/NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/FragmentNavigator.java
+++ b/NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/FragmentNavigator.java
@@ -17,10 +17,24 @@ import java.util.Map;
 
 class FragmentNavigator extends SceneNavigator {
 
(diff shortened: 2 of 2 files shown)


NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/FragmentNavigator.java
```diff
@@ -17,10 +17,24 @@
...
+        }
+
+        return null;
+    }
+
     @Override
     void navigateBack(int currentCrumb, int crumb, Activity activity, NavigationStackView stack) {
         FragmentManager fragmentManager = getFragmentManager(stack, activity);;
-        SceneFragment fragment = (SceneFragment) fragmentManager.getFragments().get(fragmentManager.getFragments().size() - 1);
+        SceneFragment fragment = getLastSceneFragment(fragmentManager);
```
grahammendick: We don't need the `getLastSceneFragment` method because all fragments are tagged with their place in the stack. So we can get the last fragment using this `oldKey` property which holds the tag of the current fragment.
```java
SceneFragment fragment = (SceneFragment) fragmentManager.findFragmentByTag(oldKey);
```
Can you test that, as well as fixing the original problem, the change works in the zoom sample because its shared element transitions rely on this line of code. There are 2 test cases to check:
1. Select a color, press back and check the shared element transition runs
2. Select a color, then change the color on the details page, press back and check the shared element transition runs.

melkayam92: Thanks for the comment.

I will check it and let you know

melkayam92:
I checked and it works good!

.gitignore
```diff
@@ -7,4 +7,6 @@ node_modules
 !Navigation*/src/node_modules
 !Navigation*/test/node_modules
 .DS_Store
-
+.project
+org.eclipse.buildship.core.prefs
+.classpath
```
grahammendick: I'm reluctant to take these changes. Could you explain how they help you, please?

melkayam92:
It's vscode auto generated staff.. I can remove it.

grahammendick:
Yes, please. Sorry for being so fussy!

melkayam92:
fixed gitignore

grahammendick:
Thanks for the bug and the PR!

grahammendick:
Your fix is live! [I thanked you in the release notes](https://github.com/grahammendick/navigation/releases/tag/v6.1.2-NavigationReactNative)

PR #323, opened 2019-11-08
Added Search Bar Component on Android

grahammendick:
The `SearchBarIOS` component is now a `SearchBar` component! On iOS it still renders to a `UISearchController`. On Android it renders to a `SearchView` action view.

Although the `SearchBar` is a child of the `NavigationBar` component, on Android it's rendered outside of the Navigation Bar so that the search results can take up the whole scene. Added a 'search' prop to the `BarButton` component because the search is triggered by a menu option press. This search `BarButton` isn’t rendered on iOS.


mpiannucci:
This is awesome. I cant wait to play around with it

grahammendick:
Thanks @mpiannucci . I've really enjoyed the challenge of translating the iOS components to Android!

Issue #325, opened 2019-11-10
Question: Set active tab to original scene on stack on tapping a tab item


arunmenon1975:
I think this qualifies more as a question than an issue since i may perhaps be missing some prop or setting.

TLDR;

a) press a tabitem and navigate to the corresponding screen 
b) Press an inner button here to navigate to another screen (non-tabbed perhaps)
c) Press any tabitem(could even be the same item tapped earlier) to get back to the main tab scene for that tab item
d) presented with the screen from b) while i expect to see the screen from a) 

I have this in my App file:

```
const stateNavigator = new StateNavigator([
  { key: 'mainmenu', trackCrumbTrail: true },
  { key: 'maindailyspecials', trackCrumbTrail: true },
  { key: 'mainmore', trackCrumbTrail: true },
  { key: 'mainadmin', trackCrumbTrail: true },
]);
 
const { mainmenu, maindailyspecials, mainmore, mainadmin } = stateNavigator.states;
mainmenu.renderScene = () => <MainMenu />;
maindailyspecials.renderScene = () => <MainDailySpecials />;
mainmore.renderScene = () => <MainMore />;
mainadmin.renderScene = () => <MainAdmin />;

const dailySpecialsNavigator = new StateNavigator(stateNavigator);
const mainMoreNavigator = new StateNavigator(stateNavigator);

stateNavigator.navigate('mainmenu');
dailySpecialsNavigator.navigate('maindailyspecials');
mainMoreNavigator.navigate('mainmore');

     <TabBar bottomTabs={true}>
        <TabBarItem title="Menu">
          <NavigationHandler stateNavigator={stateNavigator}>
            <NavigationStack
              primary={true}
              crumbStyle={from => from ? 'scale_in' : 'scale_out'}
              unmountStyle={from => from ? 'slide_in' : 'slide_out'} />
          </NavigationHandler>
        </TabBarItem>
        <TabBarItem
          title="Daily Specials"
          onPress={() => {
          }
          }
        >
          <NavigationHandler stateNavigator={dailySpecialsNavigator}>
            <NavigationStack
              primary={false}
              crumbStyle={from => from ? 'scale_in' : 'scale_out'}
              unmountStyle={from => from ? 'slide_in' : 'slide_out'} />
          </NavigationHandler>
        </TabBarItem>
        .....
        .... more tab items
        .....
        <TabBarItem
          title="More..."

          onPress={() => { }}>
[...]

PR #328, opened 2019-11-11
Added the horizontal scrollable feature to the tabbar for Android

arunmenon1975:
Please see #325 

Allows the `TabBar` to scroll horizontally when provided with a `scrollable` prop with a boolean value.

### DIFF
--- a/NavigationReactNative/src/TabBar.tsx
+++ b/NavigationReactNative/src/TabBar.tsx
@@ -10,6 +10,7 @@ class TabBar extends React.Component<any, any> {
     }
     static defaultProps = {
         bottomTabs: false,
+        scrollable: false
     }
     handleBack() {
         var {selectedTab} = this.state;
@@ -26,7 +27,7 @@ class TabBar extends React.Component<any, any> {
                 selectedTintColor={selectedTintColor}
                 unselectedTintColor={unselectedTintColor}
                 selectedIndicatorAtTop={bottomTabs}
-                scrollable={scrollable}
+                scrollable={scrollable?scrollable:false}
                 style={{backgroundColor: barTintColor, height: titleOnly ? 48 : 72}} />
         );
         return (
@@ -42,7 +43,6 @@ class TabBar extends React.Component<any, any> {
                     barTintColor={barTintColor}
                     selectedTintColor={selectedTintColor}
                     unselectedTintColor={unselectedTintColor}
-                    scrollable={scrollable}
                     style={styles.tabBar}>
                         <BackButton onPress={this.handleBack} />
                         {tabBarItems
--- a/NavigationReactNative/src/TabBar.tsx
+++ b/NavigationReactNative/src/TabBar.tsx
@@ -18,7 +18,7 @@ class TabBar extends React.Component<any, any> {
         return !!selectedTab;
     }
     render() {
-        var {children, barTintColor, selectedTintColor, unselectedTintColor, bottomTabs} = this.props;
+        var {children, barTintColor, selectedTintColor, unselectedTintColor, bottomTabs, scrollable} = this.props;
         var tabBarItems = React.Children.toArray(children).filter(child => !!child);
         var titleOnly = !tabBarItems.find(({props}: any) => props.title && typeof props.image === 'string');
         var tabLayout = Platform.OS === 'android' && (
@@ -26,6 +26,7 @@ class TabBar extends React.Component<any, any> {
                 selectedTintColor={selectedTintColor}
                 unselectedTintColor={unselectedTintColor}
                 selectedIndicatorAtTop={bottomTabs}
+                scrollable={scrollable}
                 style={{backgroundColor: barTintColor, height: titleOnly ? 48 : 72}} />
         );
         return (
@@ -41,6 +42,7 @@ class TabBar extends React.Component<any, any> {
                     barTintColor={barTintColor}
                     selectedTintColor={selectedTintColor}
                     unselectedTintColor={unselectedTintColor}
+                    scrollable={scrollable}
                     style={styles.tabBar}>
                         <BackButton onPress={this.handleBack} />
                         {tabBarItems
--- a/NavigationReactNative/src/TabBar.tsx
+++ b/NavigationReactNative/src/TabBar.tsx
@@ -27,7 +27,7 @@ class TabBar extends React.Component<any, any> {
                 selectedTintColor={selectedTintColor}
                 unselectedTintColor={unselectedTintColor}
                 selectedIndicatorAtTop={bottomTabs}
-                scrollable={scrollable?scrollable:false}
+                scrollable={scrollable}
                 style={{backgroundColor: barTintColor, height: titleOnly ? 48 : 72}} />
         );
         return (
--- a/NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/TabLayoutManager.java
+++ b/NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/TabLayoutManager.java
@@ -36,8 +36,9 @@ public class TabLayoutManager extends ViewGroupManager<TabLayoutView> {
     }
 
     @ReactProp(name = "scrollable")
-    public void setTabScrollable(TabLayoutView view, boolean scrollable) {
+    public void setTabScrollable(TabLayoutView view, @Nullable boolean scrollable) {
         view.setTabMode(scrollable ? TabLayout.MODE_SCROLLABLE : TabLayout.MODE_FIXED);
+        view.setScrollable(scrollable);
     }
 
     @Nonnull
--- a/NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/TabLayoutManager.java
+++ b/NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/TabLayoutManager.java
@@ -35,6 +35,11 @@ public class TabLayoutManager extends ViewGroupManager<TabLayoutView> {
         view.setSelectedTabIndicatorGravity(!selectedIndicatorAtTop ? TabLayout.INDICATOR_GRAVITY_BOTTOM : TabLayout.INDICATOR_GRAVITY_TOP);
     }
 
+    @ReactProp(name = "scrollable")
+    public void setTabScrollable(TabLayoutView view, boolean scrollable) {
+        view.setTabMode(scrollable ? TabLayout.MODE_SCROLLABLE : TabLayout.MODE_FIXED);
+    }
+
     @Nonnull
     @Override
     protected TabLayoutView createViewInstance(@Nonnull ThemedReactContext reactContext) {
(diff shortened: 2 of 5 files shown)


build/npm/navigation-react-native/navigation.react.native.d.ts
```diff
@@ -316,6 +316,10 @@ export interface TabBarProps {
      * Indicates whether the tabs should be at the bottom
      */
     bottomTabs?: boolean;
+    /**
+     * Indicates whether the tab bar can be scrolled horizontally
+     */
+    scrollable?: boolean;
```
grahammendick: Thanks for making the typings changes. Don't forget to copy them over to the [typings file in the test folder](https://github.com/grahammendick/navigation/blob/master/NavigationReactNative/test/node_modules/%40types/navigation-react-native.d.ts)

NavigationReactNative/src/TabBar.tsx
```diff
@@ -41,6 +42,7 @@ class TabBar extends React.Component<any, any> {
                     barTintColor={barTintColor}
                     selectedTintColor={selectedTintColor}
                     unselectedTintColor={unselectedTintColor}
+                    scrollable={scrollable}
```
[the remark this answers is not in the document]
arunmenon1975: ok

build/npm/navigation-react-native/navigation.react.native.d.ts
arunmenon1975: Ok

arunmenon1975:
> 1. Either set a default value of false for the scrollable prop in the TabBar component

I tried:

`scrollable={scrollable?scrollable:false}`

and default params:

`var {children, barTintColor, selectedTintColor, unselectedTintColor, bottomTabs, scrollable=false} = this.props;`

 on `NVTabLayout` but it doesn't seem to work


> 2. Or accept a nullable Boolean in the setTabScrollable in the TabLayoutManager
I tried `TabLayoutView view, @Nullable boolean scrollable` but it doesn't seem to work

I added the following changes to the Twitter sample App.js using `useState`:
  
`const [scrollable, setScrollable] = useState(true);`

and 

`<TabBar scrollable={scrollable} bottomTabs={false}>`

and

`
<TabBarItem
        title="Notifications"
        onPress={() => {setScrollable(null)}}>
...
`

It looks like only the initial set value is being utilized.

grahammendick:
Set a default prop value like with [the bottomTabs prop](https://github.com/grahammendick/navigation/blob/1a7422ed1e708e75e9c06447704fd2d9ae3e236d/NavigationReactNative/src/TabBar.tsx#L12)

Then try `setScrollable(false)` and `scrollable=scrollable || null}`

arunmenon1975:
i just tried it now. The changes:

```
static defaultProps = {
        bottomTabs: false,
        scrollable: false
    }
...
        var {children, barTintColor, selectedTintColor, unselectedTintColor, bottomTabs, scrollable} = this.props;

...
scrollable={scrollable || null}


```

Doesn't reflect. Only the initial value is being accepted.  Maybe i can try a `<button>` press vs the `tabitem` initiated `onPress` just to rule out any issues with the `onPress` not being registered or something like that

grahammendick:
The `scrollable={scrollable || null}` doesnt go in the TabBar. It goes in the Twitter sample

arunmenon1975:
No luck with `<Button title="Press me" onPress={() => {setScrollable(null)}} />` and `<Button title="Press me" onPress={() => {setScrollable(null)}} />`  as well.

arunmenon1975:
> The `scrollable={scrollable || null}` doesnt go in the TabBar. It goes in the Twitter sample

ok. I made the change. Reverted the TabBar component to `scrollable={scrollable?scrollable:false}` and have the Twitter sample as `scrollable={scrollable || null} `.

Doesn't seem to effect the change.

grahammendick:
Try undefined instead of null in the Twitter sample

arunmenon1975:
No change. Just to be double sure, i have it thus in the sample:

`const [scrollable, setScrollable] = useState(true);`

and it is then attached as `<TabBar scrollable={scrollable||undefined} ...`

and then button and tabItem onPress as `onPress={() => {setScrollable(false)}}`

grahammendick:
That looks right. And you have the default scrollable prop set to false i the TabBar component, right?
That doesn't pass false to your `setTabScrollable` on the Android side?

arunmenon1975:
Yes, its 

  ```
  static defaultProps = {
        bottomTabs: false,
        scrollable: false
    }
```

and 

`var {children, barTintColor, selectedTintColor, unselectedTintColor, bottomTabs, scrollable} = this.props;`

and

`scrollable={scrollable?scrollable:false}`

also tried with

`scrollable={scrollable}`

> That should pass false to your setTabScrollable on the Android side. Does it?

Is there any recommended way to debug the received values in the java file? I think `Log.*` (maybe `Log.d(scrollable);`) would require that i use Android Studio for building the project while i currently use Visual Studio Code and the default `react-native run-android` from its integrated terminal or the OS default terminal. I seem to have a problem building my app on Android studio since i get a bunch of errors like, for example:

`
ERROR: Unable to find module with Gradle path ':react-native-vector-icons' (needed by module 'app'.)
`

_[and a few a few more which includes all of RN Firebase dependencies, RN Google Sign In, RN Facebook SDK, some react-native-community packages etc. no problems with the Navigation package however]_
A not-so-educated guess is that it has something to do with `AndroidX` compatibility. The `react-native run-android` does the whole `jettifier` process each time it is run, but on Android Studio there isn't an equivalent. I have a `Log.d(scrollable);` in place inside the method but not sure how to view it.

**Comment by arunmenon1975** [comment]
i now have the log statement as:

`private static final String TAG = "TabLayoutManager";`

 and inside the method:

`Log.i(TAG, "TabLayoutManager scrollable " + scrollable);`

Will research a bit, but as of now still wondering how to access the logs;

**Comment by arunmenon1975** [comment]
I somehow got Android Studio to build>debug. Maybe the process requires stable net connection and i didnt have it in my earlier attempts.

Onload the Log.i:

`
2019-11-12 13:22:51.448 13687-13687/com.menuapp I/TabLayoutManager: TabLayoutManager scrollable true
`
onPress of the relevant tab item:

`
2019-11-12 13:24:19.985 13687-13687/com.menuapp I/TabLayoutManager: TabLayoutManager scrollable false
`
Subsequent presses doesn't log anything. Which is expected since nothing is changing.

But the UI still remains scrollable.

**Comment by arunmenon1975** [comment]
With a `onPress={() => {setScrollable((!scrollable))}}`, I see the events propagating on subsequent taps as well. So JS to native is fine.

**Comment by grahammendick** [comment]
Ok, the logs prove that the code is behaving as expected. I think we need to re-measure and layout to update the `TabLayout`. Add the following `setScrollable` method to the `TabLayoutView` and call it from the Manager:

```java
void setScrollable(boolean scrollable) {
    setTabMode(scrollable ? TabLayout.MODE_SCROLLABLE : TabLayout.MODE_FIXED);
    post(measureAndLayout);
}
```

arunmenon1975:
It works now! Also tested with the `<button>`

I also find that this may be useful to avoid the tab being scrolled to the start item whenever any tab item is pressed:

```
if(scrollable){
            view.setTabGravity(TabLayout.GRAVITY_FILL);
        }
        else {
            view.setTabGravity(TabLayout.GRAVITY_CENTER);
        }
```

Please do let me know if this can be added to `setTabScrollable`.

grahammendick:
Great! Can you show me with pictures what difference the tab gravity makes, please?

arunmenon1975:
I wasn't able to reproduce the issue after making the changes. Not sure how that happened, but the tab definitely scrolled to the beginning whenever any tab item was pressed. The issue gets noticed only when a tab item residing beyond the initial view was scrolled to and tapped  Attempting to reproduce the issue took longer than expected. I think it could have been some other issue totally non-related to the code (memory perhaps since i had a lot of applications open and maybe just specific to the emulator)

Anyway i now have the working code as:

```
public void setTabScrollable(TabLayoutView view, @Nullable boolean scrollable) {
        view.setTabMode(scrollable ? TabLayout.MODE_SCROLLABLE : TabLayout.MODE_FIXED);
        view.setScrollable(scrollable);
    }
```

and 

```
 public void setScrollable(boolean scrollable) {
        setTabMode(scrollable ? TabLayout.MODE_SCROLLABLE : TabLayout.MODE_FIXED);
        post(measureAndLayout);
    }

```
Please see the test tab interactions in the image below with this code in place:

out16

Here, `Tab 3` has the `onPress` which basically switches between boolean values in the state.

grahammendick:
Good work! Push your latest code when you're ready and I'll give it a review

arunmenon1975:
Pushed. Please check.

NavigationReactNative/src/TabBar.tsx
```diff
@@ -18,14 +19,15 @@ class TabBar extends React.Component<any, any> {
...
-        var {children, barTintColor, selectedTintColor, unselectedTintColor, bottomTabs} = this.props;
+        var {children, barTintColor, selectedTintColor, unselectedTintColor, bottomTabs, scrollable} = this.props;
         var tabBarItems = React.Children.toArray(children).filter(child => !!child);
         var titleOnly = !tabBarItems.find(({props}: any) => props.title && typeof props.image === 'string');
         var tabLayout = Platform.OS === 'android' && (
             <NVTabLayout
                 selectedTintColor={selectedTintColor}
                 unselectedTintColor={unselectedTintColor}
                 selectedIndicatorAtTop={bottomTabs}
+                scrollable={scrollable?scrollable:false}
```
grahammendick: Change to ```scrollable={scrollable}```. No need to check if populated because the default prop value handles this

NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/TabLayoutManager.java
```diff
@@ -35,6 +35,12 @@ public void setSelectedIndicatorAtTop(TabLayoutView view, boolean selectedIndica
         view.setSelectedTabIndicatorGravity(!selectedIndicatorAtTop ? TabLayout.INDICATOR_GRAVITY_BOTTOM : TabLayout.INDICATOR_GRAVITY_TOP);
     }
 
+    @ReactProp(name = "scrollable")
+    public void setTabScrollable(TabLayoutView view, @Nullable boolean scrollable) {
+        view.setTabMode(scrollable ? TabLayout.MODE_SCROLLABLE : TabLayout.MODE_FIXED);
```
grahammendick: We don't need to `setTabMode` here because the view takes care of that now. Also can we change the name to `setScrollable` instead of `setTabScrollable`?

NavigationReactNative/src/TabBar.tsx
arunmenon1975: Ok

NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/TabLayoutManager.java
arunmenon1975: Ok

arunmenon1975:
Made the changes, please check.

PR #332, opened 2019-11-17
Add backTitle property for iOS 

### LINKED ISSUE #326: Set Back Button Text on iOS
Is there a way to set the back button text for the navbar? A lot of times I dont want the title of the last screen, EG: 
Screen Shot 2019-11-10 at 8 03 44 PM

mpiannucci:
Fixes #326


grahammendick:
Thanks!

grahammendick:
Thanks, it's a great PR.

We need to handle the case where the user changes the `backTitle` after a `setState` just like we do for the `title` prop.  In the `didSetProps` of the `NVNavigationBarView`, if the `backTitle` has changed and then we need to update the previous navigation item. 

We can't subtract 2 from the count of view controllers because the user might be update the `backTitle` when it isn't the current scene. Here's a reliable way to find the previous view controller no matter what the current scene is.
```objective-c
NSInteger crumb = [self.reactViewController.navigationController.viewControllers indexOfObject:self.reactViewController];
UIViewController *previousController = crumb > 0 ? [self.reactViewController.navigationController.viewControllers objectAtIndex:crumb - 1] : nil;
```
We should use this logic in the `NVSceneController` too because I think `viewWillAppear` can fire when it isn't the current scene.

mpiannucci:
Thanks for the feedback!  Okay I’ll take a stab at updating that sometime in the next few days

grahammendick:
Here's a scenario that poses a problem. There are two scenes B and C. Scene B sets the 'backTitle' to "back" and Scene C doesn't set it. If you navigate from A to C then the back button says "A". If you then navigate back to A and navigate to B then it says "back". If you then navigate back to A and navigate to C again then it says "back" this time instead of "A"!

I think that's because scene B changes the `barBackButtonItem` on scene A's `navigationItem`. Then when you navigate to C the `barBackButtonItem` on scene A doesn't change.

What do you think we should do?

grahammendick:
I just worked out what we should do. In `viewWillAppear` we should always initialise the `backBarButtonItem` to nil and it works fine
```objective-c
previousViewController.navigationItem.backBarButtonItem = nil;
```

mpiannucci:
Awesome yeah thats the easiest solution

grahammendick:
The user might clear the 'backTitle' after a `setState` so we need to re-initialise it in the `NVNavigationBar`, too.

mpiannucci:
Finally had some time to finish this off. Let me know if I missed anything. 

In other news, once this is merged I want to work on making the tab view in android work more how it should in native apps. I'll give more detail once I take some screenshots and can explain in detail what I think should change.

NavigationReactNative/src/ios/NVNavigationBarView.m
```diff
@@ -18,6 +18,16 @@ - (void)didSetProps:(NSArray<NSString *> *)changedProps
     [self.reactViewController.navigationController setNavigationBarHidden:self.hidden];
     if ([changedProps containsObject:@"title"]) {
         [self.reactViewController.navigationItem setTitle:self.title];
+    } else if ([changedProps containsObject:@"backTitle"]) {
```
grahammendick: This should be `if` instead of an `else if` because you can change both the `title` and the `backTitle` props at the same time

NavigationReactNative/src/ios/NVNavigationBarView.m
```diff
@@ -18,6 +18,16 @@ - (void)didSetProps:(NSArray<NSString *> *)changedProps
...
+        NSInteger crumb = [self.reactViewController.navigationController.viewControllers indexOfObject:self.reactViewController];
+        UIViewController *previousController = crumb > 0 ? [self.reactViewController.navigationController.viewControllers objectAtIndex:crumb - 1] : nil;
+        if (previousController != nil) {
+            if (self.backTitle == nil) {
+                previousController.navigationItem.backBarButtonItem = nil;
+            } else {
+                previousController.navigationItem.backBarButtonItem = [[UIBarButtonItem alloc] initWithTitle:self.backTitle style:UIBarButtonItemStyleDone target:nil action:nil];
+            }
+        }
```
grahammendick: We don't need to check if the `previousController` is nil. I think it reads better if we always reset the `backBarButtonItem` to nil and then only change it if the `backTitle` is not nil
```objective-c
previousController.navigationItem.backBarButtonItem = nil;
if (self.backTitle != nil) {
    previousController.navigationItem.backBarButtonItem = [[UIBarButtonItem alloc] initWithTitle:self.backTitle style:UIBarButtonItemStyleDone target:nil action:nil];
}
```
If you're happy with this change then please also apply it to the NVSceneController

NavigationReactNative/src/ios/NVSceneController.m
```diff
@@ -49,6 +49,16 @@ - (void)viewWillAppear:(BOOL)animated
 
     [navigationBar updateColors];
 
+    NSInteger crumb = [self.navigationController.viewControllers indexOfObject:self];
+    UIViewController *previousViewController = crumb > 0 ? [self.navigationController.viewControllers objectAtIndex:crumb - 1] : nil;
```
grahammendick: Change from `previousViewController` to `previousController` to match with the code `NVNavigationBarView`

NavigationReactNative/src/ios/NVSceneController.m
```diff
@@ -49,6 +49,16 @@ - (void)viewWillAppear:(BOOL)animated
...
+    if (previousViewController != nil) {
+        if (navigationBar.backTitle == nil) {
+            previousViewController.navigationItem.backBarButtonItem = nil;
+        } else {
+            previousViewController.navigationItem.backBarButtonItem = [[UIBarButtonItem alloc] initWithTitle:navigationBar.backTitle style:UIBarButtonItemStyleDone target:nil action:nil];
```
grahammendick: Think we should use `UIBarButtonItemStylePlain` instead of `UIBarButtonItemStyleDone`. Same applies to `NVNavigationBarView`

build/npm/navigation-react-native/navigation.react.native.d.ts
```diff
@@ -69,6 +69,10 @@ export interface NavigationBarProps {
      * The color of the title view
      */
     titleColor?: string;
+    /**
+     * The title for the back button when this scene is showing
```
grahammendick: Change the comment to "The title for the back button" because there's no need for the "when scene is showing"

mpiannucci:
All resolved! Thanks!

grahammendick:
Thanks, the changes look great but there are some files showing up that shouldn't be, e.g, lock files plist etc. Can you take a look, please?

mpiannucci:
Ah man my bad. Ill go through them shortly

mpiannucci:
Sorry about that!! should good now!

grahammendick:
Thanks for the great PR and all your hard work

mpiannucci:
My pleasure! Love working on this

PR #341, opened 2019-12-31
Matched TabBar bottom tabs to Material design guide

### LINKED ISSUE #339: Use BottomNavigationView instead of top tabs for BottomTabs on Android
Currently bottom tabs are hacked on using a `TabLayout` instead of using the Material `BottomNavigationView`. There are a few important differences between the two and I will try to demonstrate with the twitter example app and the real work Twitter app on Android. 

On Android bottom tab bars should follow [the material design spec](https://material.io/components/bottom-navigation/#usage). They do not use a swipeable fragment pager, but rely on the button to transition between tabs. This may seem small but it shows when you compare the current way this library works with the Official Twitter App: 

Screenshot_1577388337

As you can see they have top tabs embedded within the fragments showing on the bottom. This is how a ton of android apps work. 

Furthermore, the sizing is wrong and there is currently no way to get rid of the indicator bar on the tab view in this library.

So in total, the height is different than what a native bottom tab bar should be, there shouldnt be a line indicator by default on bottom tabs and the child views should not be swipeable to allow for their children to have top tab views. 

I am willing to work on this, but since its a pretty big change I assume you will have some preferences in how it is implemented. Let me know any feedback you have!

grahammendick:
Fixes #339

- [x] Update selected page on ViewPager when changing tabs on TabNavigationView
- [x] Update selected tab on TabNavigationView when changing pages on ViewPager
- [x] Update the TabNavigationView when dynamically adding and removing TabBarItems
- [x] Disable swiping on ViewPager if swipeable true
- [x] ~Add hideTabs prop that doesn’t render the tabs on Android (and hides them on iOS)~ [Hiding tabs didn't work](https://github.com/grahammendick/navigation/pull/341#issuecomment-570403690)
- [x] Implement the existing tintColor props
- [x] ~Implement the existing badge and badgeColor props~ getOrCreateBadge not available in material v1
- [x] Update Twitter example to use swipeable
- [x] Set TabNavigationView height (with and without images) to match Material guidelines


mpiannucci:
For some reason the tint isnt working on icons, only on the text. Not sure why though yet but I thought Id point it out.

grahammendick:
I didn't test the icon tinting. Can you look into it?

mpiannucci:
Ya I’m looking into it

mpiannucci:
Tinting of the icons works fine with bundled android Drawable resources, but does not work with images bundled by RN. 

Not sure why, but maybe its something in the ImageResolver. Just giving an update in case this gives you an idea.

grahammendick:
Just to make sure there's nothing odd with the image you could check the tinting with the built-in ImageView. Then see if it tints ok in the NavigationBar to see if it's an isolated problem

grahammendick:
You should also check if it tints correctly with the TabLayout

mpiannucci:
Did a bunch of tests: 

1. TabNavigationView with android drawables and blue tint
drawable_bottom_nav

2. TabLayout with android drawables and blue tint
drawable_tabs

3. TabLayout with `image={require('./home.png')}` etc and blue tint
tabs_broken

4. TabNavigationView with `image={require('./home.png')}` etc and blue tint
bottom_nav_broken

5. Regular imageview with `source={require('./home.png')} and blue tint:
tintedImageView

6. NavigationBar with `BarButton` and `image={require('./home.png')}` and blue tint:
tinted_navigation_bar

For reference, here is the icon I am using to test
image

grahammendick:
Great testing, thanks. That suggests that `setTintColor` works (NavigationBar and ImageView)  but the `setTabIconTint` doesn't (TabLayout and TabNavigationView). I'm happy to release with this bug because it's an existing bug. We can fix it separately. What do you think?

mpiannucci:
Yeah same, especially because it works fine with drawables as an alternative

grahammendick:
Actually can you try a fix, please? Call `icon.setColorFilter(tintColor, PorterDuff.Mode.SRC_IN)` when the [icon resolves](https://github.com/grahammendick/navigation/blob/65c1168cdbe3c3b9c2ace7641c8ba6500a426166/NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/TabBarView.java#L62). If it works it might be worth putting it in?

grahammendick:
Hmmm, that won't work for the TabLayout because of the swiping. It's a trickier problem than I first thought. We'll release without it

mpiannucci:
The tinting works with that call but then you have to sync up the colors manually instead of it being automatic like it should.

grahammendick:
Do you know anything about Fresco, the library React Native uses for loading images on Android?

grahammendick:
Hiding the tabs didn’t work. Imagine that you’re on the Home scene and you tap a Tweet. At that point we need to hide the tabs so that they’re not visible on the Tweet scene. But the tabs are outside the stack so hiding them hides them on both the Home scene and the Tweet scene. So the tabs disappear on the Home scene as the Tweet scene animates into view. That’s no good.

So the tabs need to be part of the stack. They need to be on the Home scene. This is what the Android Home scene looks like on the Twitter app.

```jsx
const Home = () => (
  <TabBar bottomTabs={true} swipeable={false}>
    <TabBarItem title="Home">
      <Home />
    </TabBarItem>
    <TabBarItem title="Notifications" >
      <Notifications />
    </TabBarItem>
  </TabBar>
)

home.renderScene = () => <Home />;
```
And the App would be the normal stack
```jsx
const App = () => (
  <NavigationHandler stateNavigator={stateNavigator}>
    <NavigationStack />
  </NavigationHandler>
)
```
I've tried this and it works. The `TabBar` doesn't care whether or not there are stacks inside the `TabBarItems`.

mpiannucci:
I don’t know anything about Fresco but I will look into it because I’m curious why those images don’t work. 

I think that new tab bar flow is perfect

grahammendick:
A bit of background on the `IconResolver` code. I have no idea what it does but it’s meant to be an exact copy of the [ToolbarAndroid component](https://github.com/facebook/react-native/commit/cbbbb455dda313cca1a1763cc5453079ce87bc73). I copied it over for the `NavigationBar` component when the `ToolbarAndroid` was deprecated. 

I couldn’t find any other examples of an image prop in the React Native codebase. Apart from the `ImageView` component, of course, but I found that hard to follow. Even on the wider web, examples of image props on Android components are thin on the ground.

mpiannucci:
So it is indeed a problem with Fresco. I think Fresco owns the image in a native thread I believe so the tab bar controller can never take over and do the tinting. 

To test, I used the very popular [Glide](https://github.com/bumptech/glide) android library to resolve the image into a drawable from the JS bundle instead and was the tint was correct.

grahammendick:
Maybe I'm using Fresco incorrectly. Is there another way to load the image using Fresco?

mpiannucci:
Maybe you could use the [ImagePipeline](https://frescolib.org/docs/using-image-pipeline.html) directly? 

From what I gather SimpleDraweeHolder is an entire View and we just want to load an image, so maybe getting rid of the drawees and using the the fresco loader allows for that to be done

grahammendick:
Sounds like a good plan. Can you give it a bash?

mpiannucci:
Yup I ill see what I can do haha

grahammendick:
Another thing you can try is to use the built-in React Native `Image` to load images instead. Inside the TabBar component you could create Image children and set the source props from the image props. Then on the native client you can extract the Drawable and set it onto the Tab
```java
setIcon(((ImageView) child).getDrawable());
```

mpiannucci:
Using the ImagePipeline works great. I will have to clean up a bunch of stuff and make sure that the width and height props still work correctly but I think thats the best solution for now

grahammendick:
Crikey, that's v cool!

grahammendick:
The back arrow icon in the NavigationBar isn't showing up any more. Any idea why?

mpiannucci:
Oh that’s no good. I’ll look into it either tonight or tomorrow. Maybe I missed a callback somewhere.

mpiannucci:
I also noticed the Navigation bar button icons are slightly smaller than they should be too so I’ll check on that. I think those are solevable though and using the pipeline is much cleaner than the whole drawee hiararchy

grahammendick:
Yes, the cleaner code is a real winner.  I can understand your pipeline code but couldn't fathom the drawee stuff at all

mpiannucci:
Yeah same it took me a while of reading the documentation to finally get it sort of

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

mpiannucci:
So the arrow is actually showing up, it is just reallllyyy small because of the screen density. IN a real world situation you would have `arrow.png` alongside `arrow@2x.png` and `arrow@3x.png` then the correct size would be packaged by RN automatically and it would show as the correct size. 

If thats not satisfactory though I can look into why the size is incorrect and how it can be scaled up if needed.

grahammendick:
How did it work when we were using Drawees?
Can you attach arrow??.png that works so I can have a look at the difference?
Yes, please do more digging because I don't think we can release as it is.  Thanks!

mpiannucci:
All fixed with #344 , the old drawable class was needed still for sizing

grahammendick:
Awesome, the back arrow shows up in all its splendour. Thanks. Really great work

grahammendick:
Ok, we're feature complete now.

I've got to revisit the `onAttachedToWindow` methods to make sure I'm not over/under-adding listeners.

Can you add home and notifications tab images and tint them the Twitter blue to show off your work.

grahammendick:
I'm going to do some manual testing then, all being well, I'll merge it in. 

Then I'll work on the next release. That includes updating the [Tab Bar](https://grahammendick.github.io/navigation/documentation/native/tab-bar.html) documentation. I’m thinking of removing the "Adding Android Back Handlers" section and replacing it with “Adding a Bottom Navigation Bar on Android”. Now that the Navigation router supports the standard Android usecase of tabs only on the home scene, there’s no need for special back handling anymore.

mpiannucci:
Sounds great! Thanks for making this a reality! I had fun working on it

grahammendick:
The tinting doesn't work with TabLayout (unless it's a drawable resource). This isn't a showstopper because it never worked before anyway. Plus, using images on swipeable TabBars isn't a usual usecase. And swipeable tabs are Android-only so using a drawable resource isn't so bad.  But, do you have any idea why it doesn't work?

mpiannucci:
Im not sure... It should be working the same as the other tabs. Ill trya dn take a peek today

mpiannucci:
Yeah I see the same problem. I checked and it worked fine before wrapping the`BitmapDrawable` in the `DrawableWithIntrinsicSize`. So maybe that is what is doing it?

grahammendick:
Gotcha. I'll change the `DrawableWithIntrinsicSize` to subclass `BitmapDrawable` instead of wrapping it. Then I can override the intrinsic width and height

grahammendick:
Thanks, that worked a charm

grahammendick:
## Problem
Found a problem, Tap a tweet on the Home scene and the Home scene goes blank (apart from the bottom tabs) as it animates away. You might have to slow the animation down to notice. Change the animation duration in the scale_out.xml and slide_in.xml anim resource files and it’s clear as day.

It’s nothing to do with the TabBar itself (I can recreate the problem by using nested NavigationStacks). But it’s only relevant for the TabBar because it happens with the recommended bottom tabs configuration.

## Diagnosis
I think it’s a problem with nesting Fragments. Say we have an Outer Stack and an Inner Stack. The Outer Stack starts on Scene OuterA, say, and the Inner starts on InnerA. First we navigate to InnerB. Then we navigate to OuterB. This removes the inner stack completely because the Inner Stack only appears on OuterA. The Inner Stack gets confused and removes InnerB which makes OuterA look blank.

There’s nothing wrong with nesting Fragments. The Navigation router works fine with nested TabBars and nested NavigationStacks at the moment. The problem only happens when we remove an inner Fragment by navigating forward in the outer Fragment.

# Solution
I’m going to change the TabBar to use a vanilla PagerAdapter, instead of a  FragmentPagerAdapter, when there’s not a separate stack per tab.

grahammendick:
That's weird. I've been working on this problem for 2 days without any luck. Just wrote it up above and I seem to have stumbled on a fix?! Ignore my previous comment for now. Ill keep you posted

grahammendick:
I couldn’t be happier with the PR. Great working with you again

mpiannucci:
Thanks for all the hard work. BTW I am using this library in production now and its been awesome so far.

grahammendick:
That's so cool. What's your app called?

mpiannucci:
It’s called HopeWaves http://hopewaves.app/apps . The android version with the library is on Google Play and the iOS version is in beta cuz I have wayyyy more users on iOS. I rewrote everything in RN with this lib and i love the way it came out

grahammendick:
Was gonna try it out but it's not available in the UK. Keep me posted on how the iOS launch goes

mpiannucci:
Will do. Sorry about that. Its a regional Surf Forecast for the US so it is not worth the effort of deploying worldwide unfortunately

PR #343, opened 2020-01-03
Change IconResolver to use Fresco ImagePipeline directly

mpiannucci:
Allows for tinting icons in BottomNavigationView menus and TabLayout tabs

### DIFF
--- a/NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/IconResolver.java
+++ b/NavigationReactNative/src/android/app/src/main/java/com/navigation/reactnative/IconResolver.java
@@ -1,21 +1,26 @@
 package com.navigation.reactnative;
 
 import android.content.Context;
-import android.graphics.drawable.Animatable;
+import android.graphics.Bitmap;
+import android.graphics.drawable.BitmapDrawable;
 import android.graphics.drawable.Drawable;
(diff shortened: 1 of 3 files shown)


grahammendick:
Really great work. Thank you!

grahammendick:
Do we need to follow this advice from the ["Using the Image Pipeline Directly"](https://frescolib.org/docs/using-image-pipeline.html) guide?

> Using the image pipeline directly is challenging because of the memory usage. Drawees automatically keep track of whether or not your images need to be in memory. They will swap them out and load them back as soon as they need to be displayed. If you are using the image pipeline directly, your app must repeat this logic.

grahammendick:
And this advice too?

> If you do not keep a Java reference to a CloseableReference returned by the pipleine, the CloseableReference will get garbage collected and the underlying Bitmap may get recycled while still being used. If you do not close the CloseableReference once you are done with it, you risk memory leaks and OOMs.

mpiannucci:
Yeah I take care of that in the IconResolver. Because we don’t want Fresco to manage the bitmap, we copy the bitmap so it not owned by Fresco anymore then close the reference which cleans up the original but not the copy. 

If you don’t copy the bitmap it gets collected irregularly and the icons don’t show consistently.

PR #344, opened 2020-01-04
Fix drawable intrinsic sizing

mpiannucci:
This part was actually necessary to preserve correct size of icons


grahammendick:
Great detective work!

PR #361, opened 2020-03-14
Add asyncData support to native scenes

gwn:


grahammendick:
Hey, thanks so much for the PR!

I don’t think `asyncData` works that well in React Native. In a native app, the user expects the new screen to appear straight away even if all the data isn’t available. It’s better to show a placeholder or spinner than to delay the new screen animation. 

I might’ve overlooked your use-case so could you tell me how you plan on using `asyncData`, please?

gwn:
Thank you for your response :)
I understand what you are saying.

I am new in native development and I was just playing around. I thought it might be a good idea to use the `navigating` lifecycle hook for fetching stuff and was a bit surprised to see it not working with the native bindings. I just thought that you maybe forgot to add support.

I agree that a placeholder / spinner needs to be displayed while waiting for the network. But I didn't think using the `navigating` lifecycle hook would prevent me from doing so. Not sure yet but I'm thinking of setting a global state property, and displaying a spinner or whatever on top of everything when it's set to true.

gwn:
But maybe I should just use `componentDidMount` instead of doing that. I don't know. I think I need to tinker a bit more.

gwn:
Closed by mistake. Reopening, it's up to you what to do with this.

grahammendick:
Yes, using `componentDidMount` is a better approach.

The `navigating` function delays the navigation until the data returns so it's not appropriate for a native app where you want the scene animation to start right away. 

I'll close this, but thank you very much for the PR. If you have any more questions then please let me know.

Issue #351, opened 2020-01-29
java.lang.IllegalArgumentException No view found for id 0x2e7 (unknown) for fragment ScreenFragment{aba6d90 (389db4fd-87b3-4a02-8a21-096e57a6981d) id=0x2e7} 


gabrielbull:
I'm moving this issue out of [Several android crashes #348](https://github.com/grahammendick/navigation/issues/348) so we can focus on it. 

Basically, I am unable to reproduce this one on the simulator but it does happen quite often on release build on the device.

```
java.lang.IllegalArgumentException No view found for id 0x2e7 (unknown) for fragment ScreenFragment{aba6d90 (389db4fd-87b3-4a02-8a21-096e57a6981d) id=0x2e7} 
    FragmentManagerImpl.java:875 androidx.fragment.app.FragmentManagerImpl.moveToState
    FragmentManagerImpl.java:2100 androidx.fragment.app.FragmentManagerImpl.addAddedFragments
    FragmentManagerImpl.java:1874 androidx.fragment.app.FragmentManagerImpl.executeOpsTogether
    FragmentManagerImpl.java:1830 androidx.fragment.app.FragmentManagerImpl.removeRedundantOperationsAndExecute
    FragmentManagerImpl.java:1727 androidx.fragment.app.FragmentManagerImpl.execPendingActions
    FragmentManagerImpl.java:150 androidx.fragment.app.FragmentManagerImpl$2.run
    Handler.java:883 android.os.Handler.handleCallback
    Handler.java:100 android.os.Handler.dispatchMessage
    Looper.java:214 android.os.Looper.loop
    ActivityThread.java:7356 android.app.ActivityThread.main
    Method.java:-2 java.lang.reflect.Method.invoke
    RuntimeInit.java:492 com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run
    ZygoteInit.java:930 com.android.internal.os.ZygoteInit.main
```

grahammendick:
@emerson233  got the [same error](https://github.com/grahammendick/navigation/issues/348#issuecomment-576573172). It turned out that they were creating a new `StateNavigator` in the App component. This isn’t necessary because you clear the stack by setting `trackCrumbTrail` to false. Are you doing something similar?

grahammendick:
I can't recreate this one. I need more information:
- How often does it happen?
- Is there any pattern to when it happens?
- Do you create your own Android Activity (in Java)?
[...]

PR #378, opened 2020-04-22
Supported NavigationStack inside React Native Modal

grahammendick:
The following already works on iOS so this adds Android support. 
```jsx
<Modal>
  <NavigationHandler stateNavigator={stateNavigator}>
    <NavigationStack />
  </NavigationHandler>
</Modal>        
```
I thought it wasn’t possible because the React Native `Modal` isn’t a `DialogFragment`. But it wasn’t possible the way I was using `Fragments`. When a navigation happens I would try to find the closest ancestor Fragment and then use its ChildFragmentManager. Inside a Dialog there wasn’t an ancestor so it added it to the top level FragmentActivity. This doesn’t work because the stack FragmentContainer wasn’t a descendant view of the Activity. Even if it did work, the navigations from the Modal stack would pollute navigations from the primary stack because they both use the same FragmentManager.

Changed the NavigationStack so it creates its own Fragment in the top level FragmentActivity. This Fragment’s view is the NavigationStack ensuring that the FragmentContainer can be found. Then all navigations happen in the ChildFragmentManager of this Fragment. 

Got rid of the `NavigationBoundary` because don’t need to find the ancestor Fragment. Switched to a standard `PagerAdapter ` because don’t need the Fragment version.

The BackHandler doesn’t work inside a Modal. So created a `ModalBackHandler` component that can fire the back handlers from `onRequestClose`.
```jsx
<ModalBackHandler>
  {handleBack => (
    <Modal
      onRequestClose={() => {
        if (!handleBack())
          setModalVisible(false);
      }}>
      <NavigationHandler stateNavigator={stateNavigator}>
        <NavigationStack />
      </NavigationHandler>
    </Modal>        
  )}
</ModalBackHandler>
```


grahammendick:
I retested #351 after releasing this PR and couldn't recreate it (following the steps outlined in the issue). The reorganisation of the way Fragments are handled seems to have done the trick.

PR #380, opened 2020-04-23
Add specific type to offset for onOffsetChanged

salockhart:
With stricter type checking on, this `navigation-react-native` is throwing the following error:

```
node_modules/navigation-react-native/navigation.react.native.d.ts:95:49 - error TS7008: Member 'offset' implicitly has an 'any' type.

95     onOffsetChanged?: (e: NativeSyntheticEvent<{offset}>) => void;
```

This resolves this error by adding a type constraint to `onOffsetChanged`

Please let me know if the type is correct, or needs more work.

Thanks!

### DIFF
--- a/NavigationReactNative/test/node_modules/@types/navigation-react-native.d.ts
+++ b/NavigationReactNative/test/node_modules/@types/navigation-react-native.d.ts
@@ -92,7 +92,7 @@ export interface NavigationBarProps {
     /**
      * Handles offset change events
      */
-    onOffsetChanged?: (e: NativeSyntheticEvent<{offset}>) => void;
+    onOffsetChanged?: (e: NativeSyntheticEvent<{offset: number}>) => void;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


grahammendick:
Thanks for fixing the implicit 'any'. You got the type spot on.

Please manually copy the change over to the [navigation-react-native.d.ts in the test folder](https://github.com/grahammendick/navigation/blob/master/NavigationReactNative/test/node_modules/%40types/navigation-react-native.d.ts), too. The test folder is there so I can unit testing the typings (Thanks for reminding me to run them with 'noImplicitAny' from now on).

salockhart:
@grahammendick Thanks for pointing me towards that - didn't know that was even there!

Let me know if there are any other changes you would like.

Thanks!

grahammendick:
No problem. You copied the type over as a string instead of number. When you fix that I'll merge it in

salockhart:
Most not be all here today - should be okay now!

PR #394, opened 2020-05-08
Applied tint and title colors with UINavigationBarAppearance on iOS 13

grahammendick:
@mpiannucci This fixes the problems we had when the navigation bar has different `barTint` colors between scenes. Remember them? The color change animations are smooth now


mpiannucci:
Awesome!!!

PR #397, opened 2020-05-15
Add prop for setting the background color of the search bar

### LINKED ISSUE #396: NavigationBar tintColor doesn't apply to BarButton nor SearchBar on Android
I am trying to theme my app, and I want to set the navigation header tint. This works fine with `barTintColor` and `titleColor`, but when it comes to the BarButton and SearchBar components, they remain un-themed.

Here's an example branch, check out the zoom sample: https://github.com/salockhart/navigation/tree/bar-tint-example

The gist of it is:

```
<NavigationBar
  largeTitle={true}
  title="Colors"
  barTintColor="#f00"
  tintColor="#fff"
  titleColor="#fff">
```

results in:

Screenshot 2020-05-14 at 16 02 18
Screenshot 2020-05-14 at 16 02 26

Is there a different prop that I should be using to style these?

salockhart:
Closes #396 

I wasn't sure the exact place to update the type - it looked like SearchBar.tsx didn't declare its own props.

Let me know if there is anything you would like changed.

On another note - is there any documentation I should update around using `styles.xml` to style things?

### DIFF
--- a/NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/SearchBarView.java
+++ b/NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/SearchBarView.java
@@ -2,10 +2,13 @@ package com.navigation.reactnative;
 
 import android.content.Context;
 import android.view.MenuItem;
+import android.view.View;
 import android.view.ViewGroup;
 import android.view.inputmethod.InputMethodManager;
+import android.widget.ImageButton;
 
 import androidx.appcompat.widget.SearchView;
+import androidx.appcompat.widget.Toolbar;
 
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.ReactContext;
@@ -40,6 +43,19 @@ public class SearchBarView extends ReactViewGroup {
                 ReactContext reactContext = (ReactContext) getContext();
                 reactContext.getJSModule(RCTEventEmitter.class).receiveEvent(getId(),"onChangeText", event);
                 return false;
+            }
+        });
+        searchView.setOnQueryTextFocusChangeListener(new OnFocusChangeListener() {
+            @Override
+            public void onFocusChange(View v, boolean hasFocus) {
+                if (hasFocus) {
+                    NavigationBarView navigationBarView = getNavigationBarView();
+                    if (navigationBarView != null) {
+                        Toolbar toolbarView = (Toolbar) searchView.getParent();
+                        ImageButton arrowButton = (ImageButton) toolbarView.getChildAt(1);
+                        arrowButton.setColorFilter(navigationBarView.getTintColor());
+                    }
+                }
             }
         });
         onOffsetChangedListener = new AppBarLayout.OnOffsetChangedListener() {
--- a/NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/SearchBarView.java
+++ b/NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/SearchBarView.java
@@ -22,7 +22,7 @@ public class SearchBarView extends ReactViewGroup {
     int nativeEventCount;
     int mostRecentEventCount;
     int originalTextColor;
-    int originalPlaceholderColor;
+    int originalTintColor;
 
     public SearchBarView(Context context) {
         super(context);
@@ -53,7 +53,7 @@ public class SearchBarView extends ReactViewGroup {
         
         SearchView.SearchAutoComplete searchAutoComplete = getAutoComplete();
         originalTextColor = searchAutoComplete.getCurrentTextColor();
-        originalPlaceholderColor = searchAutoComplete.getCurrentHintTextColor();
+        originalTintColor = searchAutoComplete.getCurrentHintTextColor();
     }
 
     void setQuery(String query) {
@@ -62,10 +62,10 @@ public class SearchBarView extends ReactViewGroup {
             searchView.setQuery(query, true);
     }
 
-    void setBackgroundColor(Integer backgroundColor) {
+    void setBarTintColor(Integer barTintColor) {
         SearchView.SearchAutoComplete searchAutoComplete = getAutoComplete();
-        if (backgroundColor != null) {
-            searchAutoComplete.setBackgroundColor(backgroundColor);
+        if (barTintColor != null) {
+            searchAutoComplete.setBackgroundColor(barTintColor);
         } else {
             searchAutoComplete.setBackground(null);
         }
@@ -80,12 +80,12 @@ public class SearchBarView extends ReactViewGroup {
         }
     }
 
-    public void setPlaceholderColor(Integer placeholderColor) {
+    public void setTintColor(Integer tintColor) {
         SearchView.SearchAutoComplete searchAutoComplete = getAutoComplete();
-        if (placeholderColor != null) {
-            searchAutoComplete.setHintTextColor(placeholderColor);
+        if (tintColor != null) {
+            searchAutoComplete.setHintTextColor(tintColor);
         } else {
-            searchAutoComplete.setHintTextColor(originalPlaceholderColor);
+            searchAutoComplete.setHintTextColor(originalTintColor);
         }
     }
 
--- a/NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/SearchBarView.java
+++ b/NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/SearchBarView.java
@@ -51,9 +51,8 @@ public class SearchBarView extends ReactViewGroup {
                 if (hasFocus) {
                     NavigationBarView navigationBarView = getNavigationBarView();
                     if (navigationBarView != null) {
-                        Toolbar toolbarView = (Toolbar) searchView.getParent();
-                        ImageButton arrowButton = (ImageButton) toolbarView.getChildAt(1);
-                        arrowButton.setColorFilter(navigationBarView.getTintColor());
+                        ToolbarView toolbarView = (ToolbarView) searchView.getParent();
+                        toolbarView.setCollapseSearchButton((ImageButton) toolbarView.getChildAt(1));
                     }
                 }
             }
--- a/NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/SearchBarView.java
+++ b/NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/SearchBarView.java
@@ -8,7 +8,6 @@ import android.view.inputmethod.InputMethodManager;
 import android.widget.ImageButton;
 
 import androidx.appcompat.widget.SearchView;
-import androidx.appcompat.widget.Toolbar;
 
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.ReactContext;
@@ -49,11 +48,9 @@ public class SearchBarView extends ReactViewGroup {
             @Override
             public void onFocusChange(View v, boolean hasFocus) {
                 if (hasFocus) {
-                    NavigationBarView navigationBarView = getNavigationBarView();
-                    if (navigationBarView != null) {
-                        ToolbarView toolbarView = (ToolbarView) searchView.getParent();
+                    ToolbarView toolbarView = (ToolbarView) searchView.getParent();
+                    if (toolbarView.getChildAt(1) instanceof ImageButton)
                         toolbarView.setCollapseSearchButton((ImageButton) toolbarView.getChildAt(1));
-                    }
                 }
             }
         });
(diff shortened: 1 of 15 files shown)


grahammendick:
Thanks for the PR.

I’m not sure how useful it is to only change the background color of the search box. If we change it to green, for example, won’t we also want to change the search text and icons to white? Could you investigate how tricky it is to programmatically change the color of the search text and images?

Don’t forget about applying these props to the search on iOS, too.

salockhart:
Let me look into it, I think I have some time this afternoon and dig into it a bit more.

salockhart:
@grahammendick I've updated it with `textColor` and `placeholderColor` props and got those working on Android.

I have the `backgroundColor` prop working on iOS.

Getting the `textColor` and `placeholderColor` to work on iOS is proving tricky since it's on a text field buried in the UISearchBar. Hopefully I can come up with a solution in the next couple days.

salockhart:
Alright, got the iOS text color. Pretty well stuck in on the `placeholderColor`. If you've got any tips I'd be happy to try something new.

```
UITextField *searchTextField = [self findTextField:self.searchController.searchBar];
    [searchTextField setAttributedPlaceholder:[[NSAttributedString alloc] initWithString:@"blah" attributes:@{
        NSForegroundColorAttributeName:placeholderColor
    }]];
```

doesn't seem to work, and neither does the `appearanceWhenContainedInInstancesOfClasses` method.

grahammendick:
I'm surprised the `appearanceWhenContained` in works where you've put it. This Appearance api is the way to style native iOS apps but the calls should go in AppDelegate so they run once at start up. Also, the Appearance api are static calls so they apply to all SearchBars in all screens, not just the current one.

I found [something on stackoverflow that might help](https://stackoverflow.com/a/12411183/1310468) but it looks like a bit of a hack.

Have you had any luck with tinting the images for the search bar?

grahammendick:
Another idea is to [loop through the subviews of the search text field until you find the placeholder UILabel](https://stackoverflow.com/a/44329209/1310468). You might need to delay running this code until didMoveToWindow?  I think this approach should work for the images on both Android and iOS

salockhart:
I've been trying to avoid private API, which is probably why I'm having a slower time getting it working. The looping should behave similarly to the recursive search I already have - I'll give everything another try today and see how it goes.

salockhart:
@grahammendick Thanks for the tip, updating the colors in `layoutSubviews` seems to have solved it for both the text and the placeholder. I'm happy with those implementations now.

One new thing that I've noticed.. the icon/buttons. 
Screenshot 2020-05-18 at 09 22 31
Screenshot 2020-05-18 at 09 22 36

(apologies for my gaudy testing colors)

What color should those be? Be default they appear to have the same color as the placeholder, and they don't change from that when I start typing. Should we allow them to be controlled with a separate prop? Or is coloring them the same as the placeholder sufficient?

grahammendick:
Nice, you’ve made great progress!

It might be helpful if we think of the `SearchBar` component as just the text input. The search-related images that fall outside of the text input, like the search button and the close arrow on Android, we can think of as being part of the `NavigationBar` component.

We can then add new `barTintColor` and `tintColor` props to the `SearchBar` component to handle the colors (just like we have on the `NavigationBar`). The `barTintColor` is the background color of the text input. The `tintColor` is the color of the prompt and images inside the text input. Then we can add a new `textColor` prop just to pick out the text (a bit like `titleColor` prop on the navigation bar).

The images that fall outside the text input we can color using the existing `tintColor` prop of the navigation bar.

What do you think?

salockhart:
I think that makes a lot of sense. I'll refactor my changes to match that pattern

salockhart:
@grahammendick I think that pretty well covers it from my side. Let me know if you want me to change anything.

grahammendick:
Thanks. Have you done all the images on Android?

There's the back arrow that closes the search bar - we should color that using the tintColor of the navigation bar. And there's the x image inside the search bar that appears when you type - we should color that using the tintColor of the search bar.

salockhart:
Ah, no I don't think I got to those yet. I will take a peek

salockhart:
Ugh. This is going to be a little trickier than I thought.

Let's start with the "search" icon (not sure about the "x" you mentioned, I don't see that in the zoom example app, is there something I can set in the theme?).

Screenshot 2020-05-18 at 12 25 13

That icon is actually embedded in the string, from SearchView:

```
    private CharSequence getDecoratedHint(CharSequence hintText) {
        // If the field is always expanded or we don't have a search hint icon,
        // then don't add the search icon to the hint.
        if (!mIconifiedByDefault || mSearchHintIcon == null) {
            return hintText;
        }

        final int textSize = (int) (mSearchSrcTextView.getTextSize() * 1.25);
        mSearchHintIcon.setBounds(0, 0, textSize, textSize);

        final SpannableStringBuilder ssb = new SpannableStringBuilder("   ");
        ssb.setSpan(new ImageSpan(mSearchHintIcon), 1, 2, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
        ssb.append(hintText);
        return ssb;
    }
```

As far as I can tell, the only way to change that color is going to be through the `styles.xml`.

The back arrow button - that one is a standard ActionBar button as far as I can tell. I'm not sure how that gets set since our Toolbar isn't getting set as an ActionBar, and so it has no concept of the back button. Maybe the SearchBar transforms it to be an ActionBar under the hood. Again, I think this may need to be themed in the `styles.xml` since I'm not sure it's something we can set in the Toolbar.

I'll admit my Android experience is fairly low, so if you have any suggestions I'm all ears.

grahammendick:
Ok, let’s take a step back and have a look back at your original issue. You tinted the navigation bar and so you want a way to set the search bar back to white. Let’s solve that instead of trying to tint all the different search items. If we solve that then it will work for 99% of people. I don’t think anybody really wants to tint the search bar green with red text!

For the search bar to work with a tinted navigation bar, we need to only tint these two items:
- Tint the search text background white on iOS and Android so that the text is visible
- Tint the Android back arrow with the navigation bar tint color so that icon is visible 

You’ve proved we can tint the search text background white.

I’ve found out how to tint the arrow on Android. We listen for when the search is visible and then we can find the arrow image button and tint it. 

```java
searchView.setOnQueryTextFocusChangeListener(new OnFocusChangeListener() {
    @Override
    public void onFocusChange(View v, boolean hasFocus) {
        if (hasFocus) {
            ToolbarView toolbarView = (ToolbarView) searchView.getParent();
            ImageButton arrowButton = (ImageButton) toolbarView.getChildAt(1);
            arrowButton.setColorFilter(Color.WHITE);
        }
    }
});
```

What do you think to this new approach?

salockhart:
You're right - I don't think we need much of this, in fact. I will leave the `barTintColor` prop and we can remove the others.

That looks like a good solution for the back button.

I should have some time in the morning tomorrow to look at this - hopefully I will get good results.

grahammendick:
Do we even need the `barTintColor` prop? Couldn't we automatically tint the background white if the user tints the navigation bar?

salockhart:
I'd be hesitant to do that just in case the navigation bar was tinted with a light color and they still wanted to maintain the dark text color

grahammendick:
They could set a semi-transparent color, e.g., rgba(255, 0, 0, 0.2) and so not need or want to change the search text background. Ok, let's keep the `barTintColor` prop

salockhart:
@grahammendick I've updated it with what we discussed. I wasn't sure the best way of retrieving the `tintColor` prop from the navigation bar, so please let me know if there is a better way.

grahammendick:
Great work! Yeah, tinting the collapse search button (arrow button) is tricky. It might be better if we add a `collapseSearchButton` field to `ToolbarView`.

```java
private ImageButton collapseSearchButton;

void setCollapseSearchButton(ImageButton collapseSearchButton) {
  this.collapseSearchButton = collapseSearchButton;
  if (tintColor != null)
    collapseSearchButton.setColorFilter(tintColor);
  else
    collapseSearchButton.clearColorFilter();
}
```
Then we can update it from the `ToolbarView` whenever the `tintColor` changes. It works even if the search bar view is already open.
```java
void setTintColor(Integer tintColor) {
  …
  if (collapseSearchButton != null) {
    if (tintColor != null)
      collapseSearchButton.setColorFilter(tintColor);
    else
      collapseSearchButton.clearColorFilter();
  }
}
```
Here’s how to pass the button from the `SearchBarView` to the `ToolbarView`
```java
searchView.setOnQueryTextFocusChangeListener(new OnFocusChangeListener() {
    @Override
    public void onFocusChange(View v, boolean hasFocus) {
        if (hasFocus) {
            ToolbarView toolbarView = (ToolbarView) searchView.getParent();           
            toolbarView.setCollapseSearchButton((ImageButton) toolbarView.getChildAt(1));
        }
    }
});
```
Let me know what you think

salockhart:
I like it. Seems like an extensible pattern. I've updated the PR, please let me know your thoughts.

NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/SearchBarView.java
```diff
@@ -119,6 +143,10 @@ private NavigationBarView getNavigationBarView() {
         return null;
     }
 
+    private SearchView.SearchAutoComplete getAutoComplete() {
+        return searchView.findViewById(androidx.appcompat.R.id.search_src_text);
+    }
```
grahammendick, line 148: It's better if you delete this method and move this code directly into the `setBarTintColor` method.  I normally only create a separate method if it's called from more than one place

build/npm/navigation-react-native/navigation.react.native.d.ts
```diff
@@ -211,6 +211,10 @@ export interface SearchBarProps {
      * The search field text
      */
     text?: string;
+    /**
+     * The search field background color
+     */
+    barTintColor?: string;
```
grahammendick, line 217: Can you copy this change into [navigation-react-native.d.ts file in the test folder](https://github.com/grahammendick/navigation/blob/master/NavigationReactNative/test/node_modules/%40types/navigation-react-native.d.ts)

salockhart:
Thanks for the review! I knew I was going to forget to copy the types to the tests...

NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/ToolbarView.java
```diff
@@ -131,6 +133,12 @@ private void setTintColor(Drawable icon) {
             else
                 icon.clearColorFilter();
         }
+        if (collapseSearchButton != null) {
+            if (tintColor != null)
+                collapseSearchButton.setColorFilter(tintColor);
+            else
+                collapseSearchButton.clearColorFilter();
+        }
```
grahammendick, line 141: Thanks, the code looks so good. One last thing I hope!

Can you move this code into its own `setCollapseSearchTintColor` because it's called from two separate places.
Also, please can you call it from the `setTintColor(Integer tintColor)` method instead of the `setTintColor(Drawable icon)` method. **Make sure you merge in the latest master before you make this change because I [edited this setTint method when I tinted the menu items the other day](https://github.com/grahammendick/navigation/pull/400/files#diff-8993ef8d38c6f978f1483270fcaa06c5R125)**.. 
```java
void setTintColor(Integer tintColor) {
    this.tintColor = tintColor;
    setTintColor(getNavigationIcon());
    setTintColor(getLogo());
    setTintColor(getOverflowIcon());
    setCollapseSearchTintColor()
    ...
}
```

salockhart:
🤞

NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/ToolbarView.java
```diff
@@ -202,6 +205,19 @@ void setOnSearchListener(OnSearchListener onSearchListener) {
...
+        setCollapseSearchTintColor(tintColor);
+    }
+
+
+    void setCollapseSearchTintColor(Integer tintColor) {
+        if (tintColor != null)
+            collapseSearchButton.setColorFilter(tintColor);
+        else
+            collapseSearchButton.clearColorFilter();
+    }
```
grahammendick, line 219: Nearly there!
You should remove the `tintColor` parameter because the `tintColor` is a property of the class 
Also, you're missing the check that `collapseSearchButton` is not null

salockhart: done!

grahammendick:
i enjoyed working through the solution with you and I’m very happy with how this PR ended up

salockhart:
> i enjoyed working through the solution with you and I’m very happy with how this PR ended up

As am I! Thank you for your patience in getting this in.

PR #410, opened 2020-06-02
feat(NavigationReactNative): added show/hide tab bar on ios

gabrielbull:
On iOS, it is sometimes important to show/hide the tab bar, for instance, navigating to a chat screen, the chat screen will have no tab bar to allow for the message bar to be there instead.

### DIFF
--- a/NavigationReactNative/src/TabBar.tsx
+++ b/NavigationReactNative/src/TabBar.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { requireNativeComponent, Platform, StyleSheet, View } from 'react-native';
+import { requireNativeComponent, Platform, StyleSheet, View, UIManager, findNodeHandle } from 'react-native';
 import BackButton from './BackButton';
 
 class TabBar extends React.Component<any, any> {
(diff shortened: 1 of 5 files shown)


grahammendick:
Hey, thanks for the PR and the interesting feature.

Just so I can understand the use-case better, I’m going to propose two ways you can do it already and you can tell me why they don’t work.

The first way is to display the chat in a modal.

The second way is to display the `TabBar` inside a `NavigationStack`. So if we have two tabs, for example, then we need three stacks. The outer one and one for each tab. Then, when you navigate to the chat screen you do this in the outer stack.

The good thing about this is that it works on Android and iOS. The not so good thing is the iOS title goes a bit weird when swiping back from the chat to the tabs. It's because the navigation bar is on the UINavigationController on the chat screen but on the UITabBarController on the previous screen.

```jsx
const outerNavigator =  new StateNavigator([
  {key: 'home'},
  {key: 'chat'},
]);
stateNavigator.navigate('home');

const App= () => (
  <NavigationHandler stateNavigator={outerNavigator}>
    <NavigationStack />
  </NavigationHandler>
);

const {home} = stateNavigator.scenes;
home.renderScene = () => (
  <>
    <NavigationBar hidden={true} />
    <TabBar primary={true}>
      <TabBarItem title="Tab one">
        <NavigationHandler stateNavigator={tabOneNavigator}>
          <NavigationStack />
        </NavigationHandler>
      </TabBarItem>
      <TabBarItem title="Tab one">
        <NavigationHandler stateNavigator={tabTwoNavigator}>
          <NavigationStack />
        </NavigationHandler>
      </TabBarItem>
    </TabBar>
  </>
);
```

gabrielbull:
Chat in a modal defeats the purpose of the Master/Detail paradigm. The second one could work but would mess the navigation bar and would look "hacky". On Android this is not needed because when navigating to a screen, it already has no tab bar as the tab bar remains on the first level only.

There might be more use cases to hiding the tab bar on iOS than just a chat screen too.

grahammendick:
Agreed. 

The tab bar animation isn’t smooth with your solution. I didn’t run your code exactly. I used `[tabBarController.tabBar setHidden:true]`, but it’s the same idea. When you navigate from  screen A to the chat screen the tab bar disappears immediately. The problem is more obvious if you swipe back from chat to A because the tab bar doesn’t know whether it should show or hide!

The `hidesBottomBarWhenPushed` property of the `UIViewController` is better suited to the chat use-case. This associates the tab bar with the each `UIViewController` instead of the `UITabBarConroller`. So when you navigate from A to chat the tab bar sticks to screen A and disappears with it.  It works when navigating back, too. When you swipe back the tab bar appears with screen A.

grahammendick:
I’m happy to work on this if you’re busy. I’ve got a good idea of how this could work and I’ll spec it out if you want to work on it. Let me know either way

grahammendick:
If I don't hear from you over the weekend then I'll assume you're happy for me to work on this

grahammendick:
I've implemented this in #417. Thanks for letting me know about it

I added a `hidesTabBar` prop to the `NavigationStack`. For example, if there are 'home’ and 'chat’ scenes then you can hide the tab bar on the 'chat' scene.
```jsx
<NavigationStack hidesTabBar={(state) => state === chat} />
```

PR #411, opened 2020-06-02
fix(Navigation): fixed missing type definitions with typescript

### LINKED ISSUE #373: Cannot find name 'Location' when compiling React Native Typescript project
When I run `tsc` in my Typescript React Native project, I get the following error:

```
node_modules/navigation/navigation.d.ts:182:45 - error TS2304: Cannot find name 'Location'.

182     getUrl(hrefElement: HTMLAnchorElement | Location): string;
                                                ~~~~~~~~

node_modules/navigation/navigation.d.ts:230:45 - error TS2304: Cannot find name 'Location'.

230     getUrl(hrefElement: HTMLAnchorElement | Location): string;
                                                ~~~~~~~~

node_modules/navigation/navigation.d.ts:277:45 - error TS2304: Cannot find name 'Location'.

277     getUrl(hrefElement: HTMLAnchorElement | Location): string;
                                                ~~~~~~~~
```

Here are my `tsconfig` options:

```
"compilerOptions": {
    "allowJs": true,
    "allowSyntheticDefaultImports": true,
    "esModuleInterop": true,
    "isolatedModules": true,
    "jsx": "react",
    "lib": ["es6", "es2017", "es2019"],
    "moduleResolution": "node",
    "noEmit": true,
    "strict": true,
    "target": "esnext"
  },
```

If I add `dom` as a lib here, the error goes away - but then I get more errors since it's not a DOM project

gabrielbull:
Typescript is not compiling because of missing Location type

Fixes #373

### DIFF
--- a/NavigationReactNative/test/node_modules/@types/navigation-react-native.d.ts
+++ b/NavigationReactNative/test/node_modules/@types/navigation-react-native.d.ts
@@ -2,6 +2,10 @@ import { Component, Context, ReactNode } from 'react';
 import { BackHandler, ImageRequireSource, ImageURISource, NativeSyntheticEvent, StyleProp, ViewStyle } from 'react-native';
 import { Crumb, State, StateContext } from 'navigation';
 
+declare global {
+    interface Location {}
+}
+
 /**
  * Defines the Navigation Stack Props contract
  */
--- a/build/npm/navigation-react-native/navigation.react.native.d.ts
+++ b/build/npm/navigation-react-native/navigation.react.native.d.ts
@@ -2,6 +2,10 @@ import { Component, Context, ReactNode } from 'react';
 import { BackHandler, ImageRequireSource, ImageURISource, NativeSyntheticEvent, StyleProp, ViewStyle } from 'react-native';
 import { Crumb, State, StateContext } from 'navigation';
 
+declare global {
+    interface Location {}
+}
+
 /**
  * Defines the Navigation Stack Props contract
  */


grahammendick:
Ah, @salockhart raised this before and I suggested [patching it locally](https://github.com/grahammendick/navigation/issues/373#issuecomment-614796257). I'll have a rethink

gabrielbull:
It is a bit annoying to manually fix on all projects 😆

grahammendick:
How about we patch the Location in the navigation-react-native.d.ts?
```typescript
declare global {
    interface Location {}
}
```
This is similar to how [React handles the same problem](https://github.com/grahammendick/navigation/blob/master/NavigationReactNative/test/node_modules/%40types/react/global.d.ts#L33)

gabrielbull:
Fixed the PR with your changes.

grahammendick:
Thanks but please put it in the navigation-react-native.d.ts instead of navigation.d.ts. This file lives in two places (the [build folder](https://github.com/grahammendick/navigation/blob/master/build/npm/navigation-react-native/navigation.react.native.d.ts) and [test folder](https://github.com/grahammendick/navigation/blob/master/NavigationReactNative/test/node_modules/%40types/navigation-react-native.d.ts)), so make sure you copy the change to both files.

gabrielbull:
@grahammendick Are you sure, because the issue comes from the compiled `navigation/navigation.d.ts` file?

grahammendick:
`Location` is only missing on native. On mobile and web, we're in a DOM environment so `Location` is there.

gabrielbull:
Right, but it is missing from that file, so are we going to fix it by putting in `navigation-react-native.d.ts`?

If we want to fix it from mobile/native packages, we should do something like this:

```
export interface HistoryManager<Location = {}> {
    /**
     * Gets or sets a value indicating whether to disable browser history
     */
    disabled: boolean;
    /**
     * Registers browser history event listeners
     * @param navigateHistory The history navigation event handler
     */
    init(navigateHistory: (url?: string) => void): void;
    /**
     * Adds browser history
     * @param url The current url
     * @param replace A value indicating whether to replace the current
     * @param stateContext The current StateContext
     * browser history entry
     */
    addHistory(url: string, replace: boolean, stateContext: StateContext): void;
    /**
     * Gets the current location
     */
    getCurrentUrl(): string;
    /**
     * Gets an Href from the url
     */
    getHref(url: string): string;
    /**
     * Gets a Url from the anchor or location
     */
    getUrl(hrefElement: HTMLAnchorElement | Location): string;
    /**
     * Removes browser history event listeners
     */
    stop(): void;
}
```

grahammendick:
`Location` isn't a type in the Navigation router. It's a type from the typescript dom library (lib.dom.ts). Just like `HTMLElement`, for example. So we only need to patch it in for native

gabrielbull:
Okay, I'll test it out, but I'm not sure it's going to have any effect to put it there since the issues comes from the `navigation/navigation.d.ts` file.

gabrielbull:
Okay, so it works somewhat. It works if both files are imported. But it does not work if only `navigation/navigation.d.ts` is imported, which may be the case for library makers. Let me know how you want to proceed.

grahammendick:
Thanks for checking. I get what you mean. I'm happy with it because it fixes the reported problem. It's too hard to fix a problem we haven't got yet ;)!

gabrielbull:
Alright, so for now I fixed the issue from the `navigation-react-native.d.ts` files. We'll see if it happens again.

grahammendick:
Great, thanks!

PR #412, opened 2020-06-02
feat(NavigationReactNative): added styles prop for tab bar item and b…

gabrielbull:
…adge

Added fontFamily and fontSize to tab bar items and badgeStyle prop to allow styling of badge


grahammendick:
Thanks for the PR. I’ll leave aside the badge styles for now and focus on the font of the tab bar. 

Don’t the props belong on the `TabBar` rather than the `TabBarItem`? I don’t think you’d want one tab to be a different font to another. 

We can’t assume that the props are set in any order. So if the fontFamily prop is set after the fontSize then you’re overwriting the fontSize.

There are four different kinds of TabBar. - primary and non-primary on both iOS and Android. You’re only setting the props for the primary tabs on iOS.  We must make sure that the props work in all four scenarios.

grahammendick:
Please re-open this when you can set aside a chunk of time to work through it with me

grahammendick:
@gabrielbull I've added [font props to the TabBarItem](https://github.com/grahammendick/navigation/pull/450) (`fontFamily`, `fontWeight`, `fontStyle` and `fontSize`)

Issue #418, opened 2020-06-17
Android crash when navigating with the app in the background


salockhart:
Hi - I've found a very odd bug and I was wondering if you had some tips.

In my app, we have biometric security that appears when the app is loaded. Depending on the result of that biometric authentication, we then navigate to login (when it fails) or the app (when it succeeds)

The issue appears when we display the native biometric request, and then while it is requesting we close the app. This cancels the biometric request, which counts as a failure, and we try to navigate to the login screen. That navigation happens with the app already in the background.

Doing that, I get the following crash:

```
FATAL EXCEPTION: main
Process: com.app.debug, PID: 17371
java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
	at androidx.fragment.app.FragmentManagerImpl.checkStateLoss(FragmentManagerImpl.java:1536)
	at androidx.fragment.app.FragmentManagerImpl.ensureExecReady(FragmentManagerImpl.java:1672)
	at androidx.fragment.app.FragmentManagerImpl.execSingleAction(FragmentManagerImpl.java:1692)
	at androidx.fragment.app.BackStackRecord.commitNow(BackStackRecord.java:293)
	at com.navigation.reactnative.NavigationStackView.onAfterUpdateTransaction(NavigationStackView.java:66)
	at com.navigation.reactnative.NavigationStackManager.onAfterUpdateTransaction(NavigationStackManager.java:107)
	at com.navigation.reactnative.NavigationStackManager.onAfterUpdateTransaction(NavigationStackManager.java:15)
	at com.facebook.react.uimanager.ViewManager.updateProperties(ViewManager.java:51)
	at com.facebook.react.uimanager.NativeViewHierarchyManager.createView(NativeViewHierarchyManager.java:279)
	at com.facebook.react.uimanager.UIViewOperationQueue$CreateViewOperation.execute(UIViewOperationQueue.java:180)
	at com.facebook.react.uimanager.UIViewOperationQueue$1.run(UIViewOperationQueue.java:786)
	at com.facebook.react.uimanager.UIViewOperationQueue.flushPendingBatches(UIViewOperationQueue.java:903)
	at com.facebook.react.uimanager.UIViewOperationQueue.access$2400(UIViewOperationQueue.java:43)
	at com.facebook.react.uimanager.UIViewOperationQueue$2.runGuarded(UIViewOperationQueue.java:861)
	at com.facebook.react.bridge.GuardedRunnable.run(GuardedRunnable.java:31)
[...]

PR #419, opened 2020-06-17
Update NavigationStackView transaction commit to allow state loss

salockhart:
Closes https://github.com/grahammendick/navigation/issues/418


grahammendick:
Thanks v much

Issue #441, opened 2020-08-26
Cannot anchor ActionSheetIOS to BarButton on iPad


salockhart:
Hi there,

I have a use case to display an Action Sheet upon tapping a BarButton. On iPhone, this works fine, since the Action Sheet just comes up from the bottom of the screen.

On iPad, however, the Action Sheet by default appears in the middle of the screen.

![Screen-Recording-2020-08-26-at-0](https://user-images.githubusercontent.com/6651364/91284144-6d609680-e783-11ea-90e7-531a73afda0d.gif)

There is a way to "anchor" it to a node, but that requires using `findNodeHandle` from a ref, and the BarButton is a function component. When I update the BarButton to use `React.forwardRef`, it does... something, but now the Action Sheet is anchored to the top of the screen.

![Screen-Recording-2020-08-26-at-1](https://user-images.githubusercontent.com/6651364/91284162-72254a80-e783-11ea-89eb-7c6ea2b205c5.gif)

Here is the desired behaviour, from Mail:

![IMG_13BA2A721BA5-1](https://user-images.githubusercontent.com/6651364/91284745-3a6ad280-e784-11ea-8dd5-a8373ca27841.jpeg)

Any thoughts? Seems like it could be close to working, maybe something just needs to change with the frame? I've created a demo branch of this, here: https://github.com/salockhart/navigation/tree/ipad-action-sheet

grahammendick:
Nice to hear from you. And with such an interesting one, too.

The alert controller (aka action sheet) can be positioned using a view or a barButtonItem. See the ["Specifying the Popover's Anchor Point" section](https://developer.apple.com/documentation/uikit/uipopoverpresentationcontroller?language=objc). I guess the reason why they have both options is because barButtonItems aren't views (they inherit from `NSObject` not `UIView`). Unfortunately, react native doesn't accept a barButtonItem as anchor. Even if it did, `findNodeHandle` doesn't return the barButtonItem. It always returns the wrapper `NVBarButtonView`.

Perhaps we can trick it and position this wrapper view in exactly the same place as the barButtonItem? Then it doesn't matter that we're using the view instead of the button.  Inside `NVBarButtonView` and inside `didMoveToWindow` set the view frame from the button's frame. 
```obj-c
self.frame = [self.button valueForKey:@"view"].frame;
```
I haven't tried this code so it's probably wrong. But it's a rough idea to try. Can you give this a go?

salockhart:
[...]

PR #443, opened 2020-08-26
Update BarButton to allow it to be used for anchoring an ActionSheetIOS

salockhart:
Closes https://github.com/grahammendick/navigation/issues/441

The two requisite changes were:

* Updating the BarButton to pass the ref into the native NVBarButton
* Updating the NVBarButton to have its frame match that of the child button's label

Couple of questions:

1. For the `BarButton`, I've had to add types for the ref and the props. I noticed the props weren't typed before, and they don't appear to all be simple primitives. Do you have suggestions for the types here? I just left it as `any`.
2. For the `layoutSubviews` method in `NVBarButton`, let me know what sort of error handling I should take a look at here. The issue is that if we can't find the button's view, we can't really fallback to the old "show up in the middle" behaviour, since we'd still have the ref. I considered only setting the frame if the `labelView` is non-nil, what are your thoughts?

### DIFF
--- a/NavigationReactNative/src/BarButton.tsx
+++ b/NavigationReactNative/src/BarButton.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { requireNativeComponent, Image, Platform, UIManager, View, StyleSheet } from 'react-native';
 
-const BarButton = React.forwardRef<typeof NVBarButton, any>(({image, show, search, style, children, ...props}, ref) => {
+const BarButton = React.forwardRef<any, any>(({image, show, search, style, children, ...props}, ref) => {
     var constants = (UIManager as any).getViewManagerConfig('NVNavigationBar').Constants;
     return (Platform.OS === 'android' || !search) && (
         <NVBarButton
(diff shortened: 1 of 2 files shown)


salockhart:
Here's an example of it working now, with multiple buttons. Each now anchors correctly.

Screen-Recording-2020-08-26-at-1

salockhart:
- [x]     Set image or systemImage prop of the bar button instead of the title
- [x]     Change the orientation
- [x]     Use left bar button instead of right

Screenshot 2020-08-26 at 16 34 29

Realising that I chose a poor name for one of the variables, updated.

grahammendick:
Looks and works great, thanks.

Can you change both types to be `any` for the `forwardRef` call, please? Thanks for trying to find the right types but I don’t think adding types here provides much value.
```ts
React.forwardRef<any, any>(
```
I think a button's view will always be there so I'm happy with your code in `layoutSubviews`.

grahammendick:
Thanks, great issue & PR!

salockhart:
Always a pleasure @grahammendick !

Issue #442, opened 2020-08-26
RightBar BarButton ordering is opposite on iOS vs Android


salockhart:
Hi,

When I create a `RightBar` with multiple `BarButton`s, the order is reversed on iOS compared to Android.

<img width="616" alt="Screenshot 2020-08-26 at 12 49 38" src="https://user-images.githubusercontent.com/6651364/91303198-9727b680-e79f-11ea-97f8-e2ef30dd7538.png">

This doesn't appear to impact `LeftBar`

<img width="600" alt="Screenshot 2020-08-26 at 12 50 48" src="https://user-images.githubusercontent.com/6651364/91303247-a3ac0f00-e79f-11ea-961c-0dee6ff53d3b.png">

Here's some sample code from those two screenshots, from the zoom sample:

```
<NavigationBar
  title="Color"
  barTintColor={Platform.OS === 'android' ? '#fff' : null}>
  <TitleBar style={styles.titleBar}>
    <Text style={styles.titleBarText}>Color</Text>
    <View style={{backgroundColor: color, width: 28, height: 28}}/>
  </TitleBar>
  <LeftBar>
    <BarButton title="cancel 1" show="always" onPress={() => {
      stateNavigator.navigateBack(1);
    }} />
    <BarButton title="cancel 2" show="always" onPress={() => {
      stateNavigator.navigateBack(1);
    }} />
  </LeftBar>
  <RightBar>
    <BarButton title="cancel 1" show="always" onPress={() => {
      stateNavigator.navigateBack(1);
    }} />
    <BarButton title="cancel 2" show="always" onPress={() => {
      stateNavigator.navigateBack(1);
    }} />
  </RightBar>
</NavigationBar>
```

You can see this on my branch, here: https://github.com/salockhart/navigation/tree/bar-button-order

Any thoughts? It's reasonably trivial for me to get around this, but it's a bit annoying to remember to do.

grahammendick:
Yes, that’s how buttons behave on iOS. It surprised me too. If you add bar button’s A and B on the right then B appears before A. I think it’s because the buttons bar is one long row running from left to right and you’re just adding a new button on the end each time? 

I think I should’ve changed this default behaviour though? Inside `RightBar.ts` I could reverse the child `BarButtons` array on iOS. This didn’t occur to me at the time and now it would be a breaking change. I’m thinking of doing it for the next major release. What do you reckon?

salockhart:
[...]

PR #455, opened 2020-09-28
Reverse the buttons in the RightBar on iOS

salockhart:
Fixes https://github.com/grahammendick/navigation/issues/442

On iOS, we need to reverse the buttons in the RightBar, otherwise they will appear in a different order from Android.


grahammendick:
Great job. Thanks!

PR #464, opened 2020-11-05
testID Support

### LINKED ISSUE #460: Unable to add testID to navigation components: TabBarItem, BarButton, etc
As part of our application, we are writing automation tests using Detox. For that we need to be able to query for elements so that we can tap them.

Up to this point we'd been using text or testID in order to query for elements. Text works great for querying the tab bar items, but doesn't work as well if we want to query for a navigation bar button that consists of just an icon.

Would it be possible to modify these native components to apply a testID? Is that even the recommended way of doing it? We're hesitant to rely on text in all cases since we support a number of languages, and it would be helpful for our automation to be language agnostic.

salockhart:
Closes #460 

#### Nav Bar Buttons
Not too bad - just added the new prop and applied it to the button.

#### Tab Bar Buttons
Not too bad - just added the new prop and applied it to the button.

#### Back Button
Trickier, I've added the new prop and I apply it in the `layoutSubviews` step. I found that the navigation bar still had the views from the previous screen until this point - maybe there is a better way?

#### Segmented Tab Buttons
Trickiest! I wanted some way of tying the tab title and testID together, so I refactored that prop to take an array of objects instead of strings. Thinking on it more, it would be fine to probably just have two separate arrays since they would have the same length and indexes would line up. Let me know what you think.

Because this takes in a list, I needed some way of recording which segment is meant to have which testID, since titles could conceivably be duplicated. I took a stab at a dictionary indexed by the individual views, but it feels less than ideal. Happy to hear your thoughts here too.

#### Nav Bar Buttons
Need to traverse the hierarchy a bit for all of the buttons, but it seems to work out well enough. The only holdout is that we cannot seem to support the menu items in the overflow menu.

#### Bottom Tab Bar Buttons
Reasonably straight-forward

#### Back Button
Looks like we need to wait until we set the nav image before setting the testID. Is there any other way that we would set the back button?

#### Top Tab Buttons
Reasonably straight-forward

### To-do

- [x] [iOS] nav bar buttons
- [x] [iOS] tab bar buttons
- [x] [iOS] back button
- [x] [iOS] segment tab buttons
- [x] [Android] nav bar buttons
- [x] [Android] overflow menu button
- [ ] [Android] overflow menu item buttons
- [x] [Android] bottom tab bar buttons
- [x] [Android] back button
- [x] [Android] top tab buttons
- [x] Typescript


NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/ToolbarView.java
```diff
@@ -174,13 +186,32 @@ private void setTintColor(Drawable icon) {
...
                 MenuItem menuItem = getMenu().add(Menu.NONE, Menu.NONE, i, "");
                 barButton.setMenuItem(menuItem);
+                for (int j = 0; j < getChildCount(); j++) {
+                    View child = getChildAt(j);
+                    if (child instanceof ActionMenuView) {
+                        for (int k = 0; k < ((ActionMenuView) child).getChildCount(); k++) {
+                            View subchild = ((ActionMenuView) child).getChildAt(k);
+                            if (subchild instanceof ActionMenuItemView) {
+                                subchild.setTag(barButton.getTestID());
+                            }
```
salockhart, line 211: I am having difficulty here for the reason that I cannot seem to find a good way to tie a `BarButtonView` (which stores the `testID`) to the corresponding `ActionMenuItemView`. I considered using `getText` to see if the title matches, but IIRC we are not required to have a title, right? Because we can provide an icon. Any thoughts? Ideally we just check to see if it has the same reference to the `menuItem` but `getItemData` (and all methods) from `ActionMenuItemView` are marked as restricted (not sure if this causes issues? we can annotate to ignore but that seems dicey)

NavigationReactNative/src/ios/NVSegmentedTabView.m
```diff
@@ -16,16 +19,34 @@ - (id)init
...
     [self removeAllSegments];
-    for (NSString *title in titles) {
-        [self insertSegmentWithTitle:title atIndex:self.numberOfSegments animated:NO];
+    for (NSDictionary *tab in tabs) {
+        NSUInteger index = self.numberOfSegments;
+        [self insertSegmentWithTitle:tab[@"title"] atIndex:index animated:NO];
+        UIView *segment = [self subviews][index];
+        if (segment) {
+            segmentTestIDs[[NSValue valueWithNonretainedObject:segment]] = tab[@"testID"];
+        }
```
grahammendick, line 33: I’ve got a way to set the `accessibilityIdentifier` inside the loop of `setTabs`. The trick is to use the `accessibilityElements`. These point at the `UISegments` so you can then get at the labels and set the `accessibilityIdentifiers`. I know you had problems with these identifiers getting removed which was why you had to set them later. But, surprisingly, they aren’t removed when using the `accessibilityElements`!

```obj-c
UIView *segment = ((UIView *) self.accessibilityElements[index]);
segment.subviews[0].accessibilityIdentifier = tabs[index][@"testID"];
```

NavigationReactNative/src/ios/NVNavigationBarView.m
```diff
@@ -38,6 +38,23 @@ - (void)willMoveToSuperview:(nullable UIView *)newSuperview
...
+        for (UIView *view in [navigationBar subviews]) {
+            if ([view isKindOfClass:NSClassFromString(@"_UINavigationBarContentView")]) {
+                for (UIView *child in [view subviews]) {
+                    if ([child isKindOfClass:NSClassFromString(@"_UIButtonBarButton")]) {
+                        child.accessibilityIdentifier = self.navigationTestID;
+                    }
+                }
+            }
+        }
+    }
```
grahammendick, line 55: I think you can safely move this into the `updateStyle` method, but please give it a test to make sure

grahammendick (changes requested):
Thanks, it's a really great PR 👍

NavigationReactNative/src/ios/NVSegmentedTabView.m
salockhart: Hmm I wasn't able to get this working, could be an issue with my Xcode version? I'm on 11.7/iOS 13.7 for most of this.

grahammendick: Looks good. I think I only tried in iOS 14. What goes wrong?

grahammendick: Aargh, just tried it again and it's not working. Leave it with me

salockhart: Ah okay, I just tried it on Xcode 12.2 / iOS 14.2 simulator and same thing. Let me know if you have any luck!

grahammendick: Ok, it works inside the `layoutSubviews`. You just need to store the `tabs` so they're available to the `layoutSubviews`. It's better than looping through the subviews because the `accessibilityElements` are in the same order as the tabs - so can access by index
```obj-c
- (void)layoutSubviews{
    [super layoutSubviews];
    for(int i = 0; i < self.accessibilityElements.count; i++) {
        UIView *segment = ((UIView *) self.accessibilityElements[i]);
        segment.subviews[0].accessibilityIdentifier = tabs[i][@"testID"];
    }
}
```

NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/ToolbarView.java
```diff
@@ -201,20 +223,28 @@ public boolean onMenuItemActionExpand(MenuItem item) {
...
+                        ActionMenuItemView menuItemView = (ActionMenuItemView) menu.getChildAt(j);
                         if (defaultMenuTintColor == null)
-                            defaultMenuTintColor = menuItem.getCurrentTextColor();
-                        menuItem.setTextColor(tintColor != null ? tintColor : defaultMenuTintColor);
+                            defaultMenuTintColor = menuItemView.getCurrentTextColor();
+                        menuItemView.setTextColor(tintColor != null ? tintColor : defaultMenuTintColor);
+                        if (testIDs != null) {
+                            MenuItem menuItem = ((MenuView.ItemView) menuItemView).getItemData();
+                            menuItemView.setTag(testIDs.get(menuItem));
+                        }
```
grahammendick, line 243: This is another place where we need to update the button testIDs if someone changes them after the first render (see comment above about navigation and overflow testID). Let's take a shortcut here and fix this in the `BarButton` React component instead. If we set a key that's based on the testID then when the testID changes the `BarButton` will be removed and re-added so it will go into `setMenuItems` again. For safety, let's set the key based on the testID, show and title - so that if any of them change then the setMenuItems is called. Notice below that I only set the key on android if the testID is populated 
```jsx
<NVBarButton
    key={(Platform.OS === 'ios' ? systemItem : (testID ? testID + show + (title || '') : null))}
```

grahammendick (changes requested):
Thanks for making those changes. The PR is looking great!

grahammendick (changes requested):
It's looking terrific. I think we're near the end!

build/npm/navigation-react-native/navigation.react.native.d.ts
```diff
@@ -119,6 +119,14 @@ export interface NavigationBarProps {
...
      */
     backFontSize?: number;
+    /**
+     * The testID for the back button (iOS) or the navigationImage (Android)
+     */
+    navigationTestID?: string;
+    /**
+     * The testID for the overflow menu (Android only)
+     */
+    overflowTestID?: string;
```
grahammendick, line 129: Can we go with
```
/**
 * The id of the navigation button in end-to-end tests
 */
navigationTestID?: string;
/**
 * The id of the overflow button in end-to-end tests
 */
overflowTestID?: string;
```

NavigationReactNative/src/ios/NVNavigationBarManager.m
```diff
@@ -25,6 +25,7 @@ - (UIView *)view
 RCT_EXPORT_VIEW_PROPERTY(tintColor, UIColor)
 RCT_EXPORT_VIEW_PROPERTY(titleColor, UIColor)
 RCT_EXPORT_VIEW_PROPERTY(backTitle, NSString)
+RCT_EXPORT_VIEW_PROPERTY(navigationTestID, NSString)
```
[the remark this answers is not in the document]
salockhart: I thought about exactly that when I renamed the props last - how do you see us surfacing that in JSX? Would we have two separate props? Or would we have one prop in JSX that gets renamed for iOS or Android?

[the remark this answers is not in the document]
grahammendick: Two separate props

grahammendick:
This is a legendary PR!

Just to set my mind at rest, could you tell me what level of detox testing you have done against this PR. please? For example, did you run it against your [test.e2e.js](https://github.com/salockhart/navigation/commit/96fdfc7c034eca9047e2ab8b9fa2237d2a952484#diff-de3df89abe3aff5ea0f7df4a7fc98cc582823c0d36368de46fb2bfd0429443a6) with `by.id`s instead of `by.type`s?

salockhart:
Great question - I wrote up a full Detox test harness with scenarios that both use the testID and the `by.type` queries. Here's a link: https://github.com/salockhart/navigation/compare/testID-support...salockhart:detox

grahammendick:
Test harness looks great. Congrats on such a solid PR

salockhart:
Thanks for your help in getting it all working!

grahammendick:
@salockhart this PR features on the [latest episode of React Native Radio](https://reactnativeradio.com/episodes/rnr-185-navigation-with-graham-mendick)!

salockhart:
Thanks for the shout-out @grahammendick!

PR #469, opened 2020-11-20
Added Status Bar Component

grahammendick:
The built-in React Native `StatusBar` component works in the Navigation router. But a dedicated `StatusBar` component makes sense because it’s part of the navigation bar. For example, the status bar should change in `viewWillAppear` on iOS so it matches with any navigation bar changes.

Another problem with the built-in StatusBar is that it uses the legacy API on iOS (which apparently [doesn’t work when compiling extensions?)](https://pspdfkit.com/guides/ios/current/faq/view-controller-based-status-bar-appearance/). 

The new `StatusBar` component is a child of the `NavigationBar`. It works no matter the value of iOS `UIViewControllerBasedStatusBarAppearance` setting. It uses the legacy API when false (the default for a new react native project) and it uses the modern API when true.

Credit to react-native-screens for showing how to [swizzle the childViewController implementations](https://github.com/software-mansion/react-native-screens/blob/ddf1f81d5e32c9cf76179cc9396f86a0880ba0cc/ios/UIViewController%2BRNScreens.m#L41) to use the new API. React Native’s root `UIViewController` is hit first so have to swizzle to divert the call to the `SceneController`.

```jsx
<NavigationBar>
  <StatusBar tintStyle=”light” />
</NavigationBar>
```


jacobp100:
This looks awesome! The only thing missing (for me) would be animations. So if you change from one route to another, and the status bar goes from hidden to visible (or vice-versa), or changes colour, it's nice to be able to customise how that happens. I've not had any issues with how the animations were done in the standard <StatusBar> component from react-native.

grahammendick:
Thanks for checking it out. Good to know that you’re happy with the animations built-in to the core react native component. The main problem I had with the animations was getting them to work consistently with and without the `UIViewControllerBasedStatusBarAppearance` flag. The built-in comp has two props - “animated” and “showHideTransition” - but I don’t think the new API copes well with that. I didn’t think too long about it because I was happy to drop them for the initial release.

Issue #468, opened 2020-11-18
Modal Popover not centered with UIBarButtonItem


Dwar3xwar:
This is related to https://github.com/grahammendick/navigation/pull/443. 

Seems like there is an issue with centering the Action Sheet on iPad. 

When the `image` prop is provided to the `BarButton`, it seems to be the most obvious offender. It is less noticeable when the `title` or `systemItem` is used.

A potential cause for could be found in the `NVBarButton.m` in:
```
- (void)layoutSubviews
{
    [super layoutSubviews];
    UIView *buttonView = ((UIView *) [self.button valueForKey:@"view"]);
    UIView *barView = buttonView.superview;
    UIView *labelView = buttonView.subviews.count > 0 ? buttonView.subviews[0] : buttonView;
    CGRect labelFrameInBar = [buttonView convertRect:labelView.frame toView:barView];
    self.frame = [barView convertRect:labelFrameInBar toView:nil];
}
```

According to this [StackOverflow Post](https://stackoverflow.com/a/47492213), the `convertRect` does not accurately get the frame of `UIBarButtonItem` when not created with `init(customView: UIView)` and/or when the ` UINavigationBar` has not laid out all of its subviews.

grahammendick:
Thanks for letting me know and for providing good background info, too. The Stack Overflow answer is interesting. 

Could you see whether using a custom view does improve the centering, please? It's just an investigation for now so feel free to change [how it's created](https://github.com/grahammendick/navigation/blob/eda7481ed5e12e764d8defe51ba0dff9e8026d58/NavigationReactNative/src/ios/NVBarButtonView.m#L22)

Dwar3xwar:
Hey @grahammendick 

After looking into this further today, I think I may have discovered the actual issue.

It appears that when both the `title` and `image` property of a `UIBarButtonItem` is set, the `UIPopover` seems to want to center around the `title` label. This is most likely due to the fact that the `titleLabel` is the `0` index of the `subviews` of `buttonView`. 

```
- (void)layoutSubviews
{
    [super layoutSubviews];
    UIView *buttonView = ((UIView *) [self.button valueForKey:@"view"]);

    [buttonView layoutIfNeeded];

    UIView *barView = buttonView.superview;
    UIView *labelView = buttonView.subviews.count > 0 ? buttonView.subviews[0] : buttonView;
    CGRect labelFrameInBar = [buttonView convertRect:labelView.bounds toView:barView];
[...]

PR #474, opened 2020-11-23
iOS NVBarButton move frame calculation logic into buttonPressed

ndco:
Fixes: #468
Removed `layoutSubviews` and moved frame calculation logic into `buttonPressed`


grahammendick:
Thanks! I'll credit you and Eric in the release notes

PR #505, opened 2021-01-23
[WIP] Improved types

josmithua:
This is just a showcase of how you can add typing to the state keys. If this is something that is of interest I can continue working on adding type safety to the rest of the code.

### DIFF
--- a/Navigation/test/navigation-tests.ts
+++ b/Navigation/test/navigation-tests.ts
@@ -1,17 +1,28 @@
-import { Crumb, HashHistoryManager, StateNavigator, State } from 'navigation';
+import { Crumb, HashHistoryManager, StateNavigator, State, StateInfo } from 'navigation';
 
 // History Manager
-class LogHistoryManager extends HashHistoryManager  {
-    addHistory(url: string) {
-        console.log('add history');
-        super.addHistory(url, false);
-    }
+class LogHistoryManager extends HashHistoryManager {
+  addHistory(url: string) {
+    console.log('add history');
+    super.addHistory(url, false);
+  }
 }
 
 // Configuration
-const config = [
-    { key: 'people', route: ['people/{page}', 'people/{page}/sort/{sort}'], defaults: { page: 1 }, help: 'people.htm' },
-    { key: 'person', route: 'person/{id}', trackTypes: false, defaultTypes: { id: 'number' }, trackCrumbTrail: true }
+const config: StateInfo<'people' | 'person'>[] = [
+  {
+    key: 'people',
+    route: ['people/{page}', 'people/{page}/sort/{sort}'],
+    defaults: { page: 1 },
+    help: 'people.htm',
+  },
+  {
+    key: 'person',
+    route: 'person/{id}',
+    trackTypes: false,
+    defaultTypes: { id: 'number' },
+    trackCrumbTrail: true,
+  },
 ];
 const stateNavigator = new StateNavigator(config);
 stateNavigator.configure(config, new LogHistoryManager());
@@ -27,25 +38,35 @@ const idDefaultType = person.defaultTypes.id;
 // State Controller
 people.dispose = () => {};
 people.navigating = (data, url, navigate) => {
-    navigate([]);
+  navigate([]);
 };
 people.navigated = (data, asyncData) => {};
 person.navigating = (data, url, navigate) => {
-    navigate();
+  navigate();
 };
 person.navigated = (data) => {};
-person.urlEncode = function urlEncode(state: State, key: string, val: string, queryString: boolean): string {
-    return queryString ? val.replace(/\s/g, '+') : encodeURIComponent(val);
+person.urlEncode = function urlEncode(
+  state: State,
+  key: string,
+  val: string,
+  queryString: boolean
+): string {
+  return queryString ? val.replace(/\s/g, '+') : encodeURIComponent(val);
 };
-person.urlDecode = function urlDecode(state: State, key: string, val: string, queryString: boolean): string {
-    return queryString ? val.replace(/\+/g, ' ') : decodeURIComponent(val);
+person.urlDecode = function urlDecode(
+  state: State,
+  key: string,
+  val: string,
+  queryString: boolean
+): string {
+  return queryString ? val.replace(/\+/g, ' ') : decodeURIComponent(val);
 };
 person.validate = (data: any) => data.id > 0;
 person.truncateCrumbTrail = (state: State, data: any, crumbs: Crumb[]) => crumbs;
 
 // Navigation Event
 const navigationListener = (oldState: State, state: State, data: any, asyncData: any) => {
-    stateNavigator.offNavigate(navigationListener);
+  stateNavigator.offNavigate(navigationListener);
 };
 stateNavigator.onNavigate(navigationListener);
 
@@ -73,13 +94,13 @@ link = crumb.url;
 stateNavigator.navigateLink(link, 'none', true);
 
 // Fluent Navigation
-link = stateNavigator.fluent()
-    .navigate('people')
-    .refresh()
-    .refresh({ page: 3 })
-    .navigate('person', { id: 10 })
-    .navigateBack(1)
-    .url;
+link = stateNavigator
+  .fluent()
+  .navigate('people')
+  .refresh()
+  .refresh({ page: 3 })
+  .navigate('person', { id: 10 })
+  .navigateBack(1).url;
 
 // State Context
 let state: State = stateNavigator.stateContext.state;
--- a/Navigation/test/node_modules/@types/navigation.d.ts
+++ b/Navigation/test/node_modules/@types/navigation.d.ts
@@ -1,11 +1,11 @@
 /**
  * Defines a contract a class must implement in order to configure a State
  */
-export interface StateInfo {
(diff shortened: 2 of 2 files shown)


grahammendick:
Hey, thanks for the PR! Adding type safety to the navigation is of interest and it’s really great to see your work.

I think it should be more invisible to the user though. What I mean is that users shouldn’t have to define the state keys like  `StateInfo<'people' | 'person'>[]`. Can’t we work them out from the States passed in so the user doesn’t have to do anything?

I’d be interested to see how your approach extends to navigation data, too. Could you show me an example of that, please?

Also, on the web users can navigate using the `NavigationLink` component. How do you see the type checking working there?
```jsx
<NavigationLink state="person" />
```
I think you should check out [template literal types](https://twitter.com/aleksandrasays/status/1310239891117268994). They open up new possibilities for [type checking navigation](https://www.typescriptlang.org/play?ts=4.1.0-pr-40336-88#code/PTAEBsEsCMChcgOwC4FMBOAzAhgY1aAArbrYC2AKgJ4AOqAzqAN6yiiICuZ0GAXO1x7oA3K1D1k6JAHN+EqYmmi2AcnkzQAHwHcMKuZI3bOukWJUmhW8YcX6bC6dcsZRAX3jJaBAOKpkxKSU3gCMADwUoKgAHmiIACaM6ooAfKAAvKBikTFxiaAA1qhUAPaYRCTk1HSMYgD8FUHVDADaFAC6WWz8iKgAbq6e3qB+AZXBdBFRsagJSbbSaZnZ03mMAAYAJExImBigAPoAcuSobrzbu-vNbutdoA2jgVWhEd5pYgaOzoKDsF50Eb+Z5kE5kVBTXKzfLJRYZe45GZzUBbHaIPboUBgs4XNEYw43O71LGne78Ciif7DACisVIuGQACUShw0CD6JCkTCFksxLDVtDGBRiYzULgSuh4mFYQAaBwyFJiECgdBijjoeioT6gRFrFGXdH7ADKyBIyDcwCYBvxILcbmt+1FEluSrAqvoHHAyFAAApVbh1fRIAMAJTEpigFoFUBIIFjILYsIglLtfhPcbNJPjNJuUAAMlAtMkeCZLLZ4w5TuQirYyswSGw4Ci4FQ4JQ2t1gv1eONpvQ5stDsxtpdtbdDE9yHDkejsfTCdOWaCKbTwIz3iX5Bz2qYbkpAIIAAlsPR2WFGQLkbClvcL1DkairpiTWaLVae8Pxnah6Aq7d7g0kgcAQ2o4OAmr7sMzKsqgADqkDIAAFgAao2wGVpe3KODeLBsC0hAxoghTFGUhZ0iW0HlkElZpNoKg0NgSEqKmYhsAR975PRjGISorFsA8v58d0RCYYwRSlOURb0qWMFnoyNb8WwDRSRRZaoHJKT4e0QmgD0-QYLAHhUoClEMPBSGoeA6HnowHHzNh8K4ZGF6xoy9CpqAx6nhW55pHZoBAQQDSmeZKFoQwvm6b+1gRgxSH8BeHhuC0bnabAMQ0BK3riogEigGQ2BFKZmKZOeam2Vy9kKj6IYZApnJ6qK4qStKCxyj6PrgCUuCMZAJSIAAIgwuBSDQyASgl5WhZZ1mmfQKS1ekaR9CUkDxCGKQpD6mAcIgDJ9bl5IhuSdXMGIqrIOqRE7XtyAHfQ7jwMqHCaug4DYAkQwmeVGRiHRwC8WwdGZRI9CWiDyAAJLxG4gPWPRJSg+DiNQ-EvAuOgb7imQbao7wsKw19BAHAAGvCc3TeFlblYqB6HCTITk2plNWRFABEENg0wEPQ24bMpKIdOkwATEzMEs9ZHMo1zPNoxjfMC8ZxMkwAzGLaAS+znPIxI0Po78mOWtjuN6wT-OUjleURog2B9JA0iMQQuaZAVRVqegZUwfNNU+k5Nt2w7aD8H6amLWkTlsJbJQtgAdF10ghzBIbKKASXJ09YBIQwBD0IhLLgPEoA4JA4DwP79uOz6CNI9zKO8yo6fl4HqC+6AcWIfw1fIDLdcwyoqeN7bFdoL7Yjt532u17rcsG1jJQ47MeMEyoMpj73-DCwADKvkfzybaOgCE2+GenGcBYh2fiHnnqFwx9D0GXQ-N1XAODwHlexdxE-SzrqOw3KstO4qwACz9zcMnUAyp+TGANrAJuH8x5f1AF3Hu099amDngvFApsFj-zXmg0AW8d6gGNovPWyDsDQFwPEFeJ9RBAA)

josmithua:
> I think it should be more invisible to the user though. What I mean is that users shouldn’t have to define the state keys like StateInfo<'people' | 'person'>[]. Can’t we work them out from the States passed in so the user doesn’t have to do anything?

Yes, I agree it should be invisible to the user. They _are_ invisible if you pass the config into the StateNavigator directly instead of using a variable:

```ts
new StateNavigator([...])
```
But I believe the only way to have it invisible when using a variable is to have a helper function:

```ts
```

grahammendick:
Oh, that's much better!

grahammendick:
I don’t think typing the `StateNavigator` is gonna work. You see the StateNavigator has a [configure function that updates the states](https://github.com/grahammendick/navigation/blob/8bd3d65c027cbd3f452930871d1e7f443cca3414/Navigation/src/StateNavigator.ts#L30). (an example use-case is where you don’t load all your States upfront). With typing on the `StateNavigator`, it would mean that the type of the `StateNavigator` has to update.

This doesn’t mean I think typing of the Navigation router won’t work. I just think we need to find the right place to add the types. For example, I wonder if the user could create types for each `State` that specify the navigation data. 
```js
type World {
  size: number?
}

stateNavigator.navigate<World>('world', { size: 20 })
 ```
Could we somehow hook these up to the state keys as well so that typescript checks the navigation data matches the state key? We could perhaps use these types for the configuration as well?

**Comment by grahammendick** [comment]
Your idea’s really growing on me. I now think typing the StateNavigator is gonna work. I was worried about the type changing but the user can type all the StateInfos up front even if they don’t configure some of them until later.

There are advantages to having the types separate and not derived from the StateInfos, anyway. That way we can use these types in the NavigationLink and to type the stateNavigator from the `useContext` call
```js
// typed stateNavigator
const { stateNavigator } = useContext(NavigationContext<T>)
```
I think the types should be string keys (like you have already) with typed navigation data as values.  Then we can check the types of the data when navigating
```js
// type check size value against the world key
stateNavigator.navigate(‘world’, { size: 20 ))
```

grahammendick:
Here's the kind of thing I was hoping we could do but it doesn't quite work. The getData doesn't restrict the return type to just the 'world' data. Is it possible?
```ts
type N = { [key in 'hello' | 'world']: any }
class StateNavigator<T extends N> {
    constructor() {
    }
    getData(key: keyof T): T[typeof key] {
        return {} as T[typeof key];
    }
}

type S = {
    hello: {
        id: number
    },
    world: {
        size: number
    }
}
const stateNavigator = new StateNavigator<S>()
// Can we restrict this to size?
const c = stateNavigator.getData('world').size
```

grahammendick:
This works. It uses conditional types to make sure the state key matches the navigation data.
The problem is that the StateNavigator has the conditional type hard-coded in its `navigate` function. Do you know if there's a way round this?

```ts
type StateNavigationData<StateKey> = 
    StateKey extends 'hello' ? { id: string } :
    StateKey extends 'world' ? { size: number } :
    never;


class StateNavigator {
    constructor() {        
    }
    navigate<StateKey extends string>(key: StateKey, data: StateNavigationData<StateKey>) {
        return null as any;
    }
}

const stateNavigator = new StateNavigator();
stateNavigator.navigate('world', { size: 20 })
```

josmithua:
I would suggest looking at the react-navigation types. You're doing a very similar thing, except without nesting, so it should be even easier. I haven't got this running yet but this is how I envision it working:

```ts
type StateList = {
  people: { page: number };
  person: { id: number };
};
const navigator = new StateNavigator<StateList>(...);
```

navigate method in StateNavigator would look something like this:

```ts
function navigate<TStateKey extends keyof StateList>(
  stateKey: TStateKey,
  data: StateList[TStateKey]
) {
  ...
}
```

I don't really have time to keep working on this right now, but I definitely recommend looking at react-navigation's TS code.

grahammendick:
Great idea, the react navigation types have the answer for sure. I'll close this PR but thanks for the idea in the first place and for pointing me in the right direction.

grahammendick:
Just merged in #510 - compile-time strongly-typed navigation. Thanks again
```jsx
type AppNavigation = {
  people: { page?: number, sort?: string },
  person: { id: number }
}

const stateNavigator = new StateNavigator<AppNavigation>([
  { key: 'people' },
  { key: 'person' }
]);

// TypeScript error - peeple is not a state
const { peeple } = stateNavgator.states;

// TypeScript error - id is not valid for people state
const { stateNavigator } = useContext<NavigationEvent<AppNavigation>>(NavigationContext)
stateNavigator.navigate('people', { id: 12 })

// TypeScript error - id is not a number
<NavigationLink<AppNavigation, 'person'>
  stateKey="person"
  navigationData={{ id: '12' }}>
  Bob
</NavigationLink>
```

PR #509, opened 2021-02-09
On iOS, only call setNavigationBarHidden if the hidden prop changes

salockhart:
Hi there!

I was experiencing an issue in my iOS app when deeplinking and transitioning between light and dark mode: the navigation bar that was previously hidden, would now be visible. The "nested" navigation bar then appears underneath it, like so:

Screenshot 2021-02-09 at 10 54 18

This is what we would expect:

Screenshot 2021-02-09 at 11 09 08

In our app we are following a pattern similar to the docs and the twitter sample, where the `<TabBar />` component lives within a screen in the state navigator, and uses `<NavigationBar hidden />` to control the visibility of that parent navigation bar, allowing the navigation bars of the tabs to take over. We have a slight difference in that this tab bar screen is actually second in the nav stack, hence the back button in the screenshot above.

My hypothesis was that this was some sort of race condition - the tab screen would hide the nav bar, but the parent would show it, and because this is triggered at the same time with a deeplink or color scheme change, the order would not be correct.

What I have here is a possible solution - it seems to work for my app. By only calling `setNavigationBarHidden` when the `hidden` prop actually changes, I think it removes the chance of the race condition. However, that was not all that I needed - without removing the `willMoveToSuperview`, the color scheme scenario would work but the deeplinking scenario would not.

I'd love your feedback on this solution, since I am just ripping out a lifecycle method. I also wonder if it would be better to change the `navigationController` that we act on - isn't `self.reactViewController.navigationController` the currently active controller? Maybe that could be part of the problem too.

Thanks in advance! Let me know if you have any questions.

### DIFF
--- a/NavigationReactNative/src/ios/NVNavigationBarView.m
+++ b/NavigationReactNative/src/ios/NVNavigationBarView.m
@@ -16,7 +16,9 @@
 - (void)didSetProps:(NSArray<NSString *> *)changedProps
 {
     [super didSetProps:changedProps];
-    [self.reactViewController.navigationController setNavigationBarHidden:self.hidden];
+    if ([changedProps containsObject:@"hidden"]) {
+        [self.reactViewController.navigationController setNavigationBarHidden:self.hidden];
+    }
     if ([changedProps containsObject:@"title"]) {
         [self.reactViewController.navigationItem setTitle:self.title];
     }
@@ -28,14 +30,6 @@
         }
     }
     [self updateStyle];
-}
-
-- (void)willMoveToSuperview:(nullable UIView *)newSuperview
-{
-    [super willMoveToSuperview:newSuperview];
-    if (!newSuperview) {
-        [self.reactViewController.navigationController setNavigationBarHidden:false];
-    }
 }
 
 - (void)updateStyle {
--- a/NavigationReactNative/src/ios/NVNavigationBarView.m
+++ b/NavigationReactNative/src/ios/NVNavigationBarView.m
@@ -16,7 +16,7 @@
 - (void)didSetProps:(NSArray<NSString *> *)changedProps
 {
     [super didSetProps:changedProps];
-    if ([changedProps containsObject:@"hidden"]) {
+    if (self.reactViewController == self.reactViewController.navigationController.topViewController) {
         [self.reactViewController.navigationController setNavigationBarHidden:self.hidden];
     }
     if ([changedProps containsObject:@"title"]) {


salockhart:
I also attempted to reproduce this in the twitter sample app, and I wasn't able to do it properly. I was able to manually reproduce by making a change to the `Tabs` component - when React Native hot reloaded it, the issue would reproduce

grahammendick:
Hey, thanks for the PR! I tried to recreate it by running the Twitter sample and changing the Tabs component but couldn't. Could you list the steps so I can make sure I haven't missed anything out, please?

grahammendick:
I should say that your change looks good. That code in `willMoveToSuperview` doesn't smell right. I don't think it does anything because the reactViewController is probably nil when there's no superview. Although it's obviously not nil in your case.
Any reason you checked if the hidden prop has changed?

grahammendick:
Maybe we should check it's the current/active view controller like in [the check in updateStyle](https://github.com/grahammendick/navigation/blob/8bd3d65c027cbd3f452930871d1e7f443cca3414/NavigationReactNative/src/ios/NVNavigationBarView.m#L43)?
```obj-c
if (self.reactViewController == self.reactViewController.navigationController.topViewController) {
```

grahammendick:
Ha, I just recreated it. In one screen, for example, the Timeline of the Twitter sample, I delayed setting the hidden prop by a few seconds. Then I navigated to Timeline and then to a tweet before the 4 seconds elapsed. Then waited on the tweet screen for the time to pass and the navigation bar hid on the tweet screen!
```jsx
const [ hidden, setHidden ] = useState(false)
useEffect(() => {
  setTimeout(() => {
    setHidden(true)
  }, 4000)
}

<NavigationBar
  hidden={hidden}
```
So, we do need to wrap the calls to `setHidden` in a check that it's the current active screen
```obj-c
if (self.reactViewController == self.reactViewController.navigationController.topViewController) {
```
Can you try this out and see if it fixes your problem, please? You should wrap the calls in both `didSetProps` and in `willMoveToSuperview`

salockhart:
> Any reason you checked if the hidden prop has changed?

My theory was that if you had a non-active screen with a visible bar, and an active screen with a hidden bar, props changes to the non-active screen might override the change that the active screen made (if that makes sense). I wasn't able to properly test that theory but that was the root of it.

> Ha, I just recreated it. In one screen, for example, the Timeline of the Twitter sample, I delayed setting the hidden prop by a few seconds. Then I navigated to Timeline and then to a tweet before the 4 seconds elapsed. Then waited on the tweet screen for the time to pass and the navigation bar hid on the tweet screen!

Didn't even occur to me to try that! Maybe a bit contrived but I'll take what I can get 😁 

> So, we do need to wrap the calls to setHidden in a check that it's the current active screen

Let me try it out and I'll let you know

salockhart:
Seems to work really well with the new `topViewController` check, however I tried to apply the same to the operation in `willMoveToSuperview` and I'm still getting the issue in our deeplinking scenario. I've left that deleted, but if you'd like to keep it I can continue to play around with it.

grahammendick:
I'm happy you've removed `willMoveToSuperview`. It was pretty much [the very first thing I added to the Navigation Bar](https://github.com/grahammendick/navigation/pull/253/commits/7b31c18335d5e2cb6d6213348c7858315c666aa3) and I'm sure it isn't needed anymore. I guess it was trying to reset the hidden flag when the scene was popped but this taken care of by the [viewWillAppear of the Scene Controller now](https://github.com/grahammendick/navigation/blob/72f4b4fac812c34c76824f3191606090d1f811c5/NavigationReactNative/src/ios/NVSceneController.m#L55).

Thanks again for the great PR and working through the fix with me.

salockhart:
Great news! Thank you for your help.

PR #516, opened 2021-04-27
Fix NVTitleBar not reporting its final size to RN

### LINKED ISSUE #515: Text in a custom TitleBar component does not truncate on iOS
Hi there!

We are using the `TitleBar` component to build a custom title bar in our app. The main part of this title is a `Text` component, similar to what is in the `zoom` example.

However, for our use case, this text is dynamic/user-generated and can get quite long. When that happens, we cannot truncate the text properly on iOS. From the `zoom` example:

```
<TitleBar style={styles.titleBar}>
  <Text style={styles.titleBarText} numberOfLines={1}>
    Erat curabitur elit cursus est in feugiat gravida, habitant hac
    morbi urna eget per sapien, enim dictum maximus fringilla fames
    suscipit.
  </Text>
  <View style={{backgroundColor: color, width: 28, height: 28}} />
</TitleBar>
```

My expectation is that by setting `numberOfLines={1}` we should get truncated text, as per the docs: https://reactnative.dev/docs/text#numberoflines. This behaviour is present outside of the `TitleBar`.

---

My hunch is that this has something to do about the `TitleBar` not properly reporting its layout to RN. When I set a breakpoint in the `NVTitleBarView` I can see that the frame starts quite large and then shrinks to fit the available space. Maybe RN is not receiving the newly shrunken frame?

For instance, if I add a background to just the `TitleBar`, it doesn't appear to run off the page either:

and the Layout Inspector seems to confirm that:

Screenshot 2021-04-26 at 18 34 45

As a way of hacking around it, I attempted to add an `onLayout` prop to the `TitleBar`. However, it only receives one event - and it's the original big frame, not the "shrunken frame".

---

I will dig into this some more, but please do let me know if you have any ideas!

salockhart:
Fixes #515 

When the NVTitleBar is mounted, it is given a size of the entire screen. As the rest of the navigation bar mounts, that size changes - buttons in the title will make it narrower. This makes it difficult to do things like truncate text.

This PR resolves that, by calling `setSize` on the UIManager inside `layoutSubviews`. This resolves most of the issue with layout in the title bar.

There is an outstanding issue with having multiple children within the `TitleBar` - each appears to layout as though they are the only child. This can be resolved by adding padding to the `TitleBar` if the width of some children is known ahead of time.

### DIFF
--- a/NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/TitleBarView.java
+++ b/NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/TitleBarView.java
@@ -3,6 +3,10 @@ package com.navigation.reactnative;
 import android.content.Context;
 import android.view.ViewGroup;
 
+import com.facebook.react.bridge.GuardedRunnable;
+import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.uimanager.UIManagerModule;
+
 public class TitleBarView extends ViewGroup {
     public TitleBarView(Context context) {
         super(context);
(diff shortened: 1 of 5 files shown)


grahammendick:
Thanks for the PR! I don't think the [multiple children is an issue](https://github.com/grahammendick/navigation/issues/515#issuecomment-827768745). 

Can you add a check to make sure we only call `setSize` when the bounds have changed?
```obj-c
@implementation NVTitleBarView
{
    __weak RCTBridge *_bridge;
    CGRect _lastViewFrame;
}

- (void)layoutSubviews {
    [super layoutSubviews];
    if (!CGRectEqualToRect(_lastViewFrame, self.frame)) {
        [_bridge.uiManager setSize:self.bounds.size forView:self];
        _lastViewFrame = self.frame;
    }
}
```

grahammendick:
Thanks for such a great bug and PR. Great working with you again!

salockhart:
Thanks for your help debugging! Always a pleasure

PR #556, opened 2022-01-05
Re-rendered the current Scene whenever the NavigationStack re-renders

### LINKED ISSUE #554: Inconsistent behaviour when passing props through a NavigationStack in navigation-react-native
Hi, hope you're well

My team has noticed an issue recently introduced in `navigation-react-native` version `8.6.0`. Namely, when passing props all the way down from one screen, into a modal, into a navigation stack and then finally into another screen, we are seeing inconsistent behaviour.

The following can be seen at an example branch I have, here: 

### Background

Our team sometimes has need of a navigator that shows a single screen and no more. This is used for things like rendering a navigation title bar at the top of a modal. Here's an example: https://github.com/grahammendick/navigation/blob/af36d4d43cdf05001524729c757f936235354ab2/NavigationReactNative/sample/singlescreennavigator/App.js#L41

By then passing the screen as a child of `SingleScreenNavigator` or passing a render function to the `Screen` prop, one can have easy one-off navigators without too much boilerplate: https://github.com/grahammendick/navigation/blob/af36d4d43cdf05001524729c757f936235354ab2/NavigationReactNative/sample/singlescreennavigator/App.js#L117-L126

This has advantages over a standard Navigator setup for one reason: the parent component can pass props to the child directly

### <`8.6.0`

Prior to `8.6.0`, on both iOS and Android this would work with initial props, but subsequent updates to those props would not render inside the child single screen navigator's screen.

Replicable in my own project but not in the toy example provided is if the child screen is passed using a render function, instead of as a child:

https://github.com/grahammendick/navigation/blob/af36d4d43cdf05001524729c757f936235354ab2/NavigationReactNative/sample/singlescreennavigator/App.js#L129-L140

Doing it this way, the child would be able to receive updated props, likely because the way it is rendered in the scene changes - instead of passing a rendered component into the navigator, we pass a render function. I haven't yet been able to replicate this behaviour in the example project.

<details>
<summary>Videos</summary>

https://user-images.githubusercontent.com/6651364/147356852-f6666980-9dba-4547-b63a-9b26f0aa5ea8.mp4

https://user-images.githubusercontent.com/6651364/147356867-0e518965-cd12-43d9-9ae1-68a55830c12d.mp4

</details>

### >=`8.6.0`

Funny enough, with the release of `8.6.0` and also replicable in `8.7.0`, this has changed. Props now update correctly! But only on iOS. When you pass the screen as a child it works flawlessly. Passing the screen as a function works _okay_ but it is clear that something is unmounting and remounting it each time, since the spinner is no longer visible, meaning internal state gets destroyed.

<details>
<summary>Videos</summary>

https://user-images.githubusercontent.com/6651364/147359231-46786090-ee7b-4f02-a437-0c2ae8e74880.mp4

https://user-images.githubusercontent.com/6651364/147359228-e6abf440-3538-4f7b-a80b-9a6da1ea512f.mp4

</details>

### Questions

I realise this might be not-best-practice, so if the answer is "what are you doing, of course that is weird" then I value a suggestion to replace this! That said, my main questions are:

- [ ] Is this pattern doomed to fail, and are these inconsistencies a cost of that?
- [ ] If this pattern is valid, should we look at making the two platforms work consistently? Should they match the "no nav stack" example?
- [ ] How would you approach this problem? There is a lot of value in being able to pass props to a screen without it going through a navigator, but I realise that breaks the navigation model fairly substantially.

Thank you for taking a look, I appreciate it. Let me know what you think, and we can take it from there.

grahammendick:
Fixes #554

Thanks to @salockhart for raising such an interesting issue and with so much detail, too.

Take the scenario of a parent scene that opens a `Modal` with its own `NavigationStack`. The parent scene wants to pass data to the modal scene. Although it can do this using React Context it should be possible to do this using props, too. The example below uses the `renderScene` prop on the modal navigation stack to pass props to the `renderScene` on the State.

```jsx
const Wrapper = ({state, ...props}) => state.renderScene(props);

const ParentScene = ({progress, onSubmit}) => {
  const modalNavigator = useMemo(() => {
    const navigator = new StateNavigator([{key: 'scene'}])
    const {scene} = stateNavigator.states;
    screen.renderScene = (props) => <Scene {...props} />;
    navigator.navigate('scene');
    return navigator;
  });

  return (
    <NavigationHandler stateNavigator={modalNavigator}>
      <NavigationStack renderScene={(state) => (
        <Wrapper state={state} progress={progress} onSubmit={onSubmit} />
      )} />
    </NavigationHandler>
  );
}
```

The reason this didn’t work was because the Navigation router was overzealous in preventing re-renders. It stopped the scene from re-rendering if a navigation hadn’t happened. But the scene needs to re-render because the `renderScene` of the `NavigationStack` has changed.

A quick fix was to change the [lifecycle state check from RESUMED to STARTED](https://github.com/grahammendick/navigation/blob/50ca4c96728b5584e0e5b26c58c07f0bfd806ba2/NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/SceneFragment.java#L63). Changing to STARTED ensures that the `onRest` is called when the modal scene opens. In turn, this means the `shouldComponentUpdate` returns true. On iOS the `onRest` was called correctly already which explains why #554 showed the example working on iOS and not Android.

This quick fix only works if React Suspense is available. The Freeze logic added in v8.6.0 means that `shouldComponentUpdate` returns true. Before v8.6.0 the `shouldComponentUpdate` always returns false unless there’s a navigation. This explains why #554 showed the example doesn’t work on iOS or Android in < 8.6.0.

Decided not to settle for this quick fix alone. Not (just) because React Suspense might not be available but because the Freeze logic relies on some React Native internals. So there’s no way to guarantee that future versions of React Native will work with the current Freeze implementation.

In short, re-rendering the `NavigationStack` will re-render the current (topmost) scene.

### DIFF
--- a/NavigationReactMobile/src/Freeze.tsx
+++ b/NavigationReactMobile/src/Freeze.tsx
@@ -40,7 +40,7 @@ var Suspender = ({freeze, children}) => {
 };
 
 var Freeze = ({enabled, children}) => {
-    const suspendable = typeof React.Suspense !== 'undefined' && typeof window !== 'undefined';
+    const suspendable = !!React.Suspense && typeof window !== 'undefined';
     const suspender = <Suspender freeze={enabled && suspendable}>{children}</Suspender>;
     return suspendable ? <Suspense fallback={null}>{suspender}</Suspense> : suspender;
 };
--- a/NavigationReactMobile/src/NavigationMotion.tsx
+++ b/NavigationReactMobile/src/NavigationMotion.tsx
@@ -79,9 +79,10 @@ class NavigationMotion extends React.Component<NavigationMotionProps, Navigation
                     {styles => (
                         styles.map(({data: {key, state, data}, style}) => {
                             var crumb = +key.replace(/\++$/, '');
-                            var scene = <Scene crumb={crumb} renderScene={renderScene} /> ;
+                            var rest = this.state;
+                            var scene = <Scene crumb={crumb} rest={rest} renderScene={renderScene} /> ;
                             return (
-                                <Freeze key={key} enabled={this.state.rest && crumb < this.getScenes().length - 1}>
+                                <Freeze key={key} enabled={rest && crumb < this.getScenes().length - 1}>
                                     {children(style, scene, key, crumbs.length === crumb, state, data)}
                                 </Freeze>
                             );
--- a/NavigationReactMobile/src/NavigationMotion.tsx
+++ b/NavigationReactMobile/src/NavigationMotion.tsx
@@ -79,7 +79,7 @@ class NavigationMotion extends React.Component<NavigationMotionProps, Navigation
                     {styles => (
                         styles.map(({data: {key, state, data}, style}) => {
                             var crumb = +key.replace(/\++$/, '');
-                            var rest = this.state;
+                            var {rest} = this.state;
                             var scene = <Scene crumb={crumb} rest={rest} renderScene={renderScene} /> ;
                             return (
                                 <Freeze key={key} enabled={rest && crumb < this.getScenes().length - 1}>
--- a/NavigationReactMobile/src/Props.ts
+++ b/NavigationReactMobile/src/Props.ts
@@ -59,6 +59,7 @@ interface NavigationMotionProps {
 
 interface SceneProps {
     crumb: number;
+    rest: boolean;
     renderScene?: (state: State, data: any) => ReactNode;
 }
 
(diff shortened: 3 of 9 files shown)


alexstanbury:
Just wrapping my head round all this, is this line a typo?

`screen.renderScene = (props) => <Scene {...props} />;`

Should `screen.` be `scene.` ?

grahammendick:
Oh thanks! I've corrected it

PR #562, opened 2022-01-13
Add info on custom expo development client support to docs

### LINKED ISSUE #553: Docs Suggestion: Navigation Router works with expo-dev-client and EAS build
Navigation Router seems to work without any further setup with the recent "modern" expo workflow, using a custom dev client and `eas build`. So the parts of expo that don't work are using Expo Go and `expo build` (which they want to deprecate). I think that's very important to know because "no expo support" probably throws a lot of people off.

I would suggest changing

"The Navigation router for React Native uses the underlying native APIs to provide faithful navigation on Android and iOS. Because it contains native code you have to eject from Expo or create your application using react-native init."

to something like

"The Navigation router for React Native uses the underlying native APIs to provide faithful navigation on Android and iOS. Because it contains native code you have to create a custom development build or eject when using Expo or create your application using react-native init."

and link "custom development build" to https://docs.expo.dev/development/introduction/

Those docs then make clear than you have to use EAS Build or a fully local build and cannot use classic expo build / expo go. Using just EAS Build and not custom development client will work, but is a pretty terrible workflow and not worth mentioning imho.

I would have made a PR instead of an issue, but I don't think the docs are on GitHub, are they?

karlsander:
Adding the wording suggested in #553 (with minimal changes):

"The Navigation router for React Native uses the underlying native APIs to provide faithful navigation on Android and iOS. Because it contains native code, you have to create your application using react-native init or a custom development build of Expo."

Fixes #553

### DIFF
--- a/documentation/native/setup.html
+++ b/documentation/native/setup.html
@@ -51,8 +51,8 @@
           </div>
           <div id="article">
             <h1>Setup</h1>
-            <p>
-                The Navigation router for React Native uses the underlying native APIs to provide faithful navigation on Android and iOS. Because it contains native code you have to eject from Expo or create your application using <code>react-native init</code>.
+             <p>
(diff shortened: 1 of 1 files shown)


grahammendick:
Thanks @karlsander

PR #583, opened 2022-02-10
Added labelVisibility prop to TabBar

### LINKED ISSUE #577: Android Tab layout when rendering more than 3 tabs
Hi,

I've run into an issue where, when adding more than 3 TabBarItems to a TabBar on Android, it is removing the title from inactive tabs and centring the icon. The only title visible is that of the active Tab, as you can see in the screenshot below.

Tab issue

Would it be possible to somehow disable this behaviour and force the tabs to render their full contents (as they do on iOS)?

SourceErr0r:
fixes #577 

Exposed `setLabelVisibilityMode` method in Android TabNavigationManager as `labelVisibility` on `<TabBar />` with 3 options: 

`auto`: Default behaviour
`show`: Show all labels
`hide`: Hide all labels
`selected`: Only show label for active tab


grahammendick:
Thanks for the PR and [the stacktrace](https://github.com/grahammendick/navigation/issues/577#issuecomment-1034877701). This is [where the stacktrace points](https://github.com/material-components/material-components-android/blob/4c7be527751f1cb11f7c5f55ca5eecbb41c1620c/lib/java/com/google/android/material/bottomnavigation/BottomNavigationMenuView.java#L85). The [isShifting method](https://github.com/material-components/material-components-android/blob/4c7be527751f1cb11f7c5f55ca5eecbb41c1620c/lib/java/com/google/android/material/navigation/NavigationBarMenuView.java#L766) annoyingly doesn't check for children if the mode is SELECTED.

You can work around this by delaying the setting of the visibility mode. You can create a field on the `TabNavigationView` class to hold the visibility mode. Then only set it once there are children.

Yea, prettier is pretty annoying like that. Please undo those changes and get this PR down to only your intentional changes

grahammendick:
Hmmm...I've gone off my idea. Let's stick with your original idea

NavigationReactNative/src/NavigationBar.tsx
```diff
@@ -89,4 +89,4 @@ var Container: any = ({collapse, children, ...props}) => (
 var NVNavigationBar = requireNativeComponent<any>('NVNavigationBar', null);
 var NVToolbar = requireNativeComponent<any>('NVToolbar', null);
 
-export default Animated.createAnimatedComponent(NavigationBar);
+export default Animated.createAnimatedComponent(NavigationBar);
```
grahammendick, line 92: Can you remove this change so this file doesn't show up in the diff, please? Looks like you've added a new line

SourceErr0r:
> Hmmm...I've gone off my idea. Let's stick with your original idea

Haha, I have no more time to work on this at the moment so that's good with me :)

Would we not be able to achieve the same effect on iOS by simply not setting a title, though?

NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/TabNavigationView.java
```diff
@@ -63,6 +64,7 @@ protected void onAttachedToWindow() {
             setSelectedItemId(tabBar.selectedTab);
             autoSelected = false;
             tabBar.populateTabs();
+            this.setLabelVisibilityMode(labelVisibility);
```
grahammendick, line 67: This works for the first time the tab bar renders, but what about if the user changes the `labelVisibility` later on? You can test it like this,
```jsx
const [labelVisibility, setLabelVisibility] = useState('selected');
useEffect(() => {
  setTimeout(() => {
    setLabelVisibility('hide')
  }, 1000)
), []);
<TabBar labelVisibility={labelVisibility} />
```

package-lock.json
```diff
@@ -46,7 +46,7 @@
       }
     },
     "build/npm/navigation-react-mobile": {
-      "version": "3.6.4",
+      "version": "3.7.0",
```
grahammendick, line 49: Can you undo this change please? Think a merge from master should fix it

SourceErr0r:
> Will you have the time?

Of course, but it might not be until next week as I'm away for the weekend, and need to fit this around other tasks :)

NavigationReactNative/src/TabBar.tsx
```diff
@@ -54,9 +54,20 @@ class TabBar extends React.Component<any, any> {
...
+        // Default to auto
+        var labelVisibilityVal = -1;
+
+        if(labelVisibility === "selected") {
+          labelVisibilityVal = 0;
+        } else if (labelVisibility === "show") {
+          labelVisibilityVal = 1;
+        } else if (labelVisibility === "hide") {
+          labelVisibilityVal = 2;
+        }
```
grahammendick, line 66: Instead of hard-coding these constants, you can get them from the java code like [I've done in the SearchBar](https://github.com/grahammendick/navigation/blob/9e9c0f7efc3906581f8b8a0720ff35f75305142f/NavigationReactNative/src/SearchBar.tsx#L36). You need to [export them from the view manager](https://github.com/grahammendick/navigation/blob/9e9c0f7efc3906581f8b8a0720ff35f75305142f/NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/SearchBarManager.java#L80)

build/npm/navigation-react-native/navigation.react.native.d.ts
```diff
@@ -525,6 +525,14 @@ export interface TabBarProps {
...
      */
     defaultTab?: number;
+    /**
+     * Whether to show or hide labels in the tab bar (Android Only)
+     * selected - only show on active tab
+     * auto - "show" when <4 tabs, "selected" when >3
+     * show - always show.
+     * hide - always hide (no labels)
+     */
+    labelVisibility?: "auto" | "show" | "hide" | "selected";
```
grahammendick, line 535: Can you use 'labeled', 'unlabeled' and 'selected' instead to match the java constants?

build/npm/navigation-react-native/navigation.react.native.d.ts
```diff
@@ -703,4 +711,4 @@ export function useUnloading(handler: (state: State, data: any, url: string, his
  * Registers callback for when this Scene has navigated to another Scene
  * @param handler The unloaded event handler
  */
-export function useUnloaded(handler: (state: State, data: any, stateContext: StateContext) => void) : void;
+export function useUnloaded(handler: (state: State, data: any, stateContext: StateContext) => void) : void;
```
grahammendick, line 714: Can you remove this change so this file doesn't show up in the diff, please? Looks like you've added a new line

grahammendick (commented):
Thanks again for the PR. Let me know if you have any questions about the review comments

build/npm/navigation-react-native/navigation.react.native.d.ts
SourceErr0r: Yep can do - I was trying to make the props a little more 'user friendly' but happy to keep it lined up with the base constants.

grahammendick:
>> Of course, but it might not be until next week 

Great, whenever you have the free time obviously cool!

NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/TabNavigationView.java
SourceErr0r: Good point, hadn't considered that use case

NavigationReactNative/src/TabBar.tsx
SourceErr0r: I've had a free hour today so I've been working on this, but I cannot figure out how to do as you suggest. I can't seem get the constants from any Android classes, the docs say they exist on NavigationBarView but we're overriding that with the React Native implementation which does not export those constants. Any advice on how to do this?

grahammendick: Thanks for picking this up again. You probably need to fully qualify the constants with the namespace so picks the right class. For example, 
```java
com.google.android.material.navigation.NavigationBarView.LABEL_VISIBILITY_SELECTED
```

SourceErr0r: Aha, didn't know I could do it like that; I'd tried adding it as an import at the top in an attempt to get it using the right one but this is much more direct! Thanks

NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/TabNavigationManager.java
```diff
@@ -21,6 +24,13 @@ public void setBottomTabs(TabNavigationView view, boolean bottomTabs) {
         view.bottomTabs = bottomTabs;
     }
 
+    @ReactProp(name = "labelVisibility")
+    public void setLabelVisibilityMode (TabNavigationView view, int labelVisibilityMode) {
+        view.labelVisibility = labelVisibilityMode;
+        view.setVisibility();
+    }
```
grahammendick, line 31: Here's the twitter sample with label visibility set to "auto"
Screenshot 2022-02-11 at 17 11 37

And here it is with "selected"
Screenshot 2022-02-11 at 17 10 20

You can see that Android changes the width of the tabs when label visibility is "selected". I think when there are more than 3 tabs this width change isn't noticeable. But now we're allowing "selected" with <= 3 tabs the width change isn't good.

We can get rid of this width change by disabling `itemHorizontalTranslation`.
```java
setItemHorizontalTranslationEnabled(false);
```
We only want to disable this if label visibility is "selected" and there are <= 3 tabs. The translation runs an animation when the tab is selected if there are more than 3 tabs, and we don't want to lose that. So you can add this to `TabBar.tsx`
```jsx
<TabView
  labelVisibility={labelVisibility}
  itemHorizontalTranslation={labelVisibility !== 'selected' || tabBarItems.length > 3}
```
Then on the java side you can add a corresponding `setItemHorizontalTranslation` method.

The good news is that this means we can simplify the `setLabelVisibilityMode`. We don't need the `labelVisibility` or `canSetVisibility` fields anymore because, when `itemHorizontalTranslation` is disabled, it doesn't [throw the NullPointer](https://github.com/grahammendick/navigation/issues/577#issuecomment-1034849460) anymore! 
```java
@ReactProp(name = "labelVisibility")
public void setLabelVisibilityMode(TabNavigationView view, int labelVisibilityMode) {
  view.setLabelVisibilityMode(labelVisibilityMode);
}
```

NavigationReactNative/src/TabBar.tsx
```diff
@@ -43,7 +43,7 @@ class TabBar extends React.Component<any, any> {
         return false;
     }
     render() {
-        var {children, barTintColor, selectedTintColor, unselectedTintColor, bottomTabs, scrollable, primary, scrollsToTop} = this.props;
+        var {children, labelVisibility = "auto", barTintColor, selectedTintColor, unselectedTintColor, bottomTabs, scrollable, primary, scrollsToTop} = this.props;
```
grahammendick, line 46: Please use `defaultProps` to set the default of 'auto'. That way the default value will be set no matter where we access the props from

NavigationReactNative/src/TabBar.tsx
```diff
@@ -54,9 +54,14 @@ class TabBar extends React.Component<any, any> {
         var TabBar = primary ? NVTabBar : TabBarPager;
         var TabView = primary ? NVTabNavigation : (!I18nManager.isRTL ? NVTabLayout : NVTabLayoutRTL);
         TabView = Platform.OS === 'android' ? TabView : NVSegmentedTab;
+        // Default to auto
+
```
grahammendick, line 58: Please remove the code comments

NavigationReactNative/src/TabBar.tsx
```diff
@@ -112,4 +117,4 @@ const styles = StyleSheet.create({
     },
 });
 
-export default TabBar;
+export default TabBar;
```
grahammendick, line 120: Can you the newline back so this line doesn't show up in the diff, please?

build/npm/navigation-react-native/navigation.react.native.d.ts
```diff
@@ -525,6 +525,14 @@ export interface TabBarProps {
...
      */
     defaultTab?: number;
+    /**
+     * Whether to show or hide labels in the tab bar (Android Only)
+     * selected - only show on active tab
+     * auto - "show" when <4 tabs, "selected" when >3
+     * labeled - always show
+     * unlabeled - always hide (no labels)
+     */
+    labelVisibility?: "auto" | "labeled" | "unlabeled" | "selected";
```
grahammendick, line 535: Thanks for writing a comprehensive typing comment. It's so good that it's inconsistent with my other comments ;). Can you change it to the following, please?
```ts
/**
 * Indicates how labels should be displayed in the tab bar
 */
labelVisibility?: "auto" | "labeled" | "unlabeled" | "selected";
```

grahammendick (commented):
Thanks for your latest updates! As always, let me know if you have any questions

NavigationReactNative/src/android/src/main/java/com/navigation/reactnative/TabNavigationManager.java
SourceErr0r: Unless... What if we just default to 'auto' when there are more than 3 tabs? It would have the same behaviour, without the need for the java workaround.

grahammendick: Yes, that's a great idea! Default to 'auto' when there are more than 3

build/npm/navigation-react-native/navigation.react.native.d.ts
SourceErr0r: Sorry, did that out of habit. But you're right, I should aim to keep it consistent! Fixed this

NavigationReactNative/src/TabBar.tsx
SourceErr0r: This _should_ be fixed now. I think it's something about the way my editor is handling the EoF. If I add lines, it seems to modify the final line as well. Very annoying.

NavigationReactNative/src/TabBar.tsx
```diff
@@ -54,9 +55,19 @@ class TabBar extends React.Component<any, any> {
...
+
+        if(Platform.OS === 'android'){
+            var constants = (UIManager as any).getViewManagerConfig('NVTabNavigation').Constants;
+            visibilityMode = (labelVisibility === 'selected' && tabBarItems.length > 3) ? constants.labelVisibility['auto'] : constants.labelVisibility[labelVisibility]
+        }
+
         var tabLayout = (Platform.OS === 'android' || !primary) && (
             <TabView
                 bottomTabs={bottomTabs}
+                labelVisibility={visibilityMode}
```
grahammendick, line 69: I prefer the following. What do you think?
```jsx
labelVisibility = !(labelVisibility === 'selected' && tabBarItems.length > 3) ? labelVisibility : 'auto';
<TabView
  labelVisibility={Platform.OS === 'android' ? constants.labelVisibility[labelVisibility] : null}
```

grahammendick (commented):
It's looking great. Nice brainwave you had about changing to 'auto' 🧠

grahammendick (commented):
It's looking great. Nice brainwave you had about changing to 'auto' 🧠

build/npm/navigation-react-native/navigation.react.native.d.ts
```diff
@@ -525,6 +525,10 @@ export interface TabBarProps {
      * The default selected tab index
      */
     defaultTab?: number;
+    /**
+     * Indicates how labels should be displayed in the tab bar
+     */
+    labelVisibility?: "auto" | "labeled" | "unlabeled" | "selected";
```
grahammendick, line 531: Do you think we should call it `labelVisibilityMode` instead of `labelVisibility` to match the android field? React Native sometimes use [mode to match android](https://reactnative.dev/docs/scrollview#overscrollmode-android)

NavigationReactNative/src/TabBar.tsx
SourceErr0r: I was trying to find a way to make this a bit more concise but Monday evening brain just wasn't doing it :) Will use your suggestion here. 

I would suggest assigning this to a new variable rather than overriding the input prop though, we need to keep a reference for the `itemHorizontalTranslation` prop. Unless we compare again using constants, but I feel that's overcomplicating things. What do you think?

grahammendick: Don't think we need a new variable. The `itemHorizontalTranslation` will be true when we change `selected` to `auto` so should be fine, right?

SourceErr0r: If tabs < 4 we want to keep 'selected', and will need to turn off `itemHorizontalTranslation`. I could do something like 

`constants.labelVisibility[labelVisibility] !== constants.labelVisibility['selected']`

But I feel like just keeping a reference to the original prop makes more sense here.

SourceErr0r: 🤦  I've just realised on second look that we are not setting the constant as a variable. I'll fix that now.

grahammendick:
It's looking brilliant! This line is world class
```jsx
itemHorizontalTranslation={labelVisibilityMode !== 'selected'}
```
Thanks for your hard work on this. I'll do some testing and, all being well, I'll merge it in

SourceErr0r:
No problem, thanks for the help in refining it!

In terms of testing - I'm having trouble with XCode at the moment (currently doing a fresh install as nothing else seems to be working), so while I did run through all of the Android variations I haven't been able to test whether this has any adverse effects on iOS. It shouldn't, as the props would (I think) just be ignored, but it's certainly something that needs looking at :)

grahammendick:
Thanks for letting me know! I always test on both before merging but it's v helpful to know what testing you've done

NavigationReactNative/src/TabBar.tsx
```diff
@@ -54,9 +55,14 @@ class TabBar extends React.Component<any, any> {
         var TabBar = primary ? NVTabBar : TabBarPager;
         var TabView = primary ? NVTabNavigation : (!I18nManager.isRTL ? NVTabLayout : NVTabLayoutRTL);
         TabView = Platform.OS === 'android' ? TabView : NVSegmentedTab;
+
+        var constants = Platform.OS === 'android' ? (UIManager as any).getViewManagerConfig('NVTabNavigation').Constants : null;
+        labelVisibilityMode = !(labelVisibilityMode === 'selected' && tabBarItems.length > 3) ? labelVisibilityMode : 'auto';
         var tabLayout = (Platform.OS === 'android' || !primary) && (
             <TabView
                 bottomTabs={bottomTabs}
+                labelVisibilityMode={Platform.OS === 'android' ? constants.LabelVisibility[labelVisibilityMode] : null}
```
grahammendick, line 64: Ok, you don't need another platform check then. You can use optional chaining instead
```jsx
labelVisibilityMode={constants?.LabelVisibility[labelVisibilityMode]}
```

SourceErr0r: Good spot, I'll add this in

PR #584, opened 2022-03-19
Prevented blank screen when navigations and animations overlap

grahammendick:
Take the scenario A → B. Inside A, there's a 4 second timer that navigates back to A.
```jsx
useEffect(() => {
  const timer = setInterval(() => {
    stateNavigator.navigate('A');
}, 4000);
  return () => clearInterval(timer);
})
```
Then click a button on B that navigates to A → B → C.

Time it right so that the button click happens just before the timer runs. So it first navigates back to A and then navigates to A → B → C. Then press back from C and scene B will be blank!

This happens because scenes were keyed by position, so scene B has the same key whenever it appears. When the animation back to A completes then scene B is popped. It should only pop the first scene B but it pops both B's because they share a key.

Fixed by giving scenes unique keys using a counter every that increments every navigation. So the first set of keys are 0-0 → 1-1 then when navigate back to A the counter increments and when navigate to C it increments again and gives 0-0 → 3–1 → 3-2. When animation back to A completes then scene 1-1 (first B) pops. When going back to second B it's not blank anymore because it's still there in React.

Note, that on the old MacBook Air the error happened but it didn't on the M1 MacBook. It was just too fast. So created fake slowdowns after each navigation to force the backup and then it failed on the M1 MacBook, too. For example,
```jsx
stateNavigator.navigate('A');
const date = Date.now();
let currentDate = null;
do {
  currentDate = Date.now();
} while (currentDate - date < 1000);
```

### DIFF
--- a/NavigationReactNative/src/NavigationStack.tsx
+++ b/NavigationReactNative/src/NavigationStack.tsx
@@ -32,11 +32,11 @@ class NavigationStack extends React.Component<NavigationStackProps, NavigationSt
         if (!state)
             return {keys: []};
         var prevState = prevStateNavigator && prevStateNavigator.stateContext.state;
-        var currentKeys = crumbs.concat(nextCrumb).map((_, i) => `${counter}-${i}}`);
+        var currentKeys = crumbs.concat(nextCrumb).map((_, i) => `${counter}-${i}`);
         var newKeys = currentKeys.slice(prevKeys.length);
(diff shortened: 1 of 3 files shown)


grahammendick:
Note, I rolled back this PR because [iOS relies keys being in old format](https://github.com/grahammendick/navigation/blob/8ecee5b68e99afe03658a61595b6a36b860705bb/NavigationReactNative/src/ios/NVNavigationStackView.m#L137). Will fix that and then create a new PR

PR #638, opened 2022-09-05
[Documentation] Add missing self-closing tag

burakgormek:
Just a small thing but it can be frustrating when copy&paste.


grahammendick:
Thank you 👍

PR #641, opened 2022-09-12
android: remove backgroundColor from scene root view

### LINKED ISSUE #639: Android (scene) background color
Hi! As always, it's so pleasant using your navigation library. This time a come w/ a couple of requests. This is for Android, next for iOS. I could manage to help w/ a PR if you let me know some guidelines in how to do it, eg how do you try things before release, because I've _touching_ sources under `node_modules` and obviously it's not the best approach!

Anyway, I have an app w/ dark background. I noticed on Android that there's a flash-y feeling just before all the screen is rendered, caused because the default background color for everything in RN Android is white. I tried the `underlayColor` prop but no success. After debugging with Flipper layout inspector I saw that changing the opacity of `SceneView` to zero, the background color from the stack `underlayColor` is there. First I tried to add this to. `SceneViewManager`:

```java
@ReactProp(name = "underlayColor", customType = "Color", defaultInt = Integer.MAX_VALUE)
public void setUnderlayColor(SceneView view, int underlayColor) {
    view.setBackgroundColor(underlayColor);
}
```

It doesn't work yet because I guess I need to change things on js land... btw, I preferred ask here before to check out how do you see this...

Thanks!

iamyellow:
`crumbStyle` `unmountStyle` transitions using alpha shown unwanted white color caused by hardcoded background color on Scene.

---

Fixes #639

before:

https://user-images.githubusercontent.com/1761227/189640131-d2f4e3bc-87bc-4843-9007-a90aefb4e507.mp4

after:

https://user-images.githubusercontent.com/1761227/189640004-05e14116-6f99-4a51-bf27-cea3847949fb.mp4

### DIFF
--- a/NavigationReactNative/src/Scene.tsx
+++ b/NavigationReactNative/src/Scene.tsx
@@ -167,7 +167,6 @@ var NVScene = global.nativeFabricUIManager ? require('./SceneNativeComponent').d
 
 const styles = StyleSheet.create({
     scene: {
-        backgroundColor: '#fff',
         position: 'absolute',
         top: 0, right: 0,
         bottom: 0, left: 0,


grahammendick:
@iamyellow this is a good PR but a transparent background doesn't work with large titles on iOS. Also, transparent background is a breaking change. So I'm working on a way to bring this in as a prop so can keep the white default and allow it to be changed. I'll keep you posted

PR #659, opened 2022-12-19
Supported Material Design 3 for the BottomAppBar (Android) 

grahammendick:
Updated Android material components to v1.7.0 to support [Material Design 3 for the BottomAppBar](https://m3.material.io/components/bottom-app-bar/specs). Set height to 80dp to match the Material 3 specs if the user opts in by using a `Material3` Theme.

[Material components v1.7.0](https://github.com/material-components/material-components-android/releases/tag/1.7.0-beta01) updated the `compileSdkVersion` and `targetSdkVersion` to 32. Currently, React Native 0.70 has these at 31 so not going to merge this PR yet. It looks like [React Native 0.71 bumps these to 33](https://github.com/facebook/react-native/pull/35196) which would avoid having to add a manual bump step to the setup instructions.


grahammendick:
React Native has bumped `compileSdkVersion` and `targetSdkVersion` to 33 so good to merge

PR #661, opened 2023-01-04
backIndicatorImage on iOS appearance

iamyellow:
For some reason some native image icons (Images.xcassets) were displayed as a big white rounded button. This change sets to `UIImageRenderingModeAlwaysOriginal` the image rendering mode displaying the image as expected.


iamyellow:
sure! I've attached the icon svg + before vs after. hope it helps!

back
before
after

grahammendick:
Hey, thanks again for the PR. I tested it out but it only works for svg asset images. I tried it with a png asset `backImage={{ uri: 'back' }}` and a require `backImage={require('./back.png')}` but they both show up blank. Do you know a solution that will work for all flavours?

image

grahammendick:
I've got it working without any changes. 

image

I set the "Render As" to "Original Image" in Xcode

image

iamyellow:
cool, it works :) didn't know about that option.

PR #664, opened 2023-01-21
Fix swipe back when hiding navbar

RichardLindhout:

PR #677, opened 2023-02-25
Supported swipe back when NavigationBar is hidden (iOS)

### LINKED ISSUE #477: [IOS] there shouldn't be a navigation bar if no component is provided ?
Screen Shot 2020-11-27 at 10 18 50 AM

on IOS an navigation is visible by default and you need to render `<NavigationBar hidden />` to hide it.

also it doesn't animate correctly when going from a screen with NavigationBar hidden to another with it visible.

grahammendick:
Fixes #664

@RichardLindhout is migrating from RNN to the Navigation router. Migrating all his custom navigation bars over to the Navigation router's native `NavigationBar` is too much in one go. So he wants to turn off the native navigation bar and use his own for now. But setting `hidden` to true on the `NavigationBar` prevents the swipe back gesture on iOS.

There's a difference between `setNavigationBarHidden` on `UINavigationController` and `setHidden` on the `UINavigationBar`. The first disables the swipe but the second doesn't. I'm not sure if it's official iOS behaviour (it still works on iOS 16).

If the `NavigationBar` is hidden checked for `backTitle` to decide which way to hide the navigation bar. If there's a `backTitle` used `setHidden` to keep the swipe.

```jsx
// Swipe disabled
<NavigationBar hidden />

// Swipe enabled
<NavigationBar hidden backTitle="something" />
```
The original plan was to use the presence of the `NavigationBar` to determine whether to dis/able the swipe. But the Navigation router doesn’t support a `StatusBar` without a `NavigationBar`. Couldn’t hide the navigation bar, keep the swipe and customise the status bar.

While investigating this original plan found a fast way to check for no `NavigationBar` at all using the `NSNotificationCenter`. Firing a notification that the `NavigationBar` receives avoids having to search through all descendant views.

Even though the PR moved away from this plan, kept the `NSNotificationCenter` in place. Ever since Fabric stole tags, wanted a fast way to find the `NavigationBar` (replacing `viewWithTag`). For example, it’s possible to put the `NavigationBar` at the bottom of the page (Android `CoordinatorLayout` examples can put it at the bottom) which would’ve meant trawling through all views to find it.


RichardLindhout:
It works, thank you!

Issue #691, opened 2023-04-01
Remove or Hide Flashing while Navigating [Font Family not getting reflected]


afkcodes:
Hello, Graham, thanks for this amazing navigation library was pretty simple to setup, also how can i theme the bottom navigation bar

How can i reduce flashing while navigation. I am attaching a screen-record of the same. This is not visible in some of the apps that are built natively also on react native that i know.

https://user-images.githubusercontent.com/17390257/229299259-a9912a49-4372-4d8f-b7ef-e374a38bcc95.mp4

grahammendick:
Hey Ashish, thanks for getting in touch.

You'll find the [documentation for the bottom tabs under the section titled 'Adding Top Level Navigation Tabs'](https://grahammendick.github.io/navigation/documentation/native/tab-bar.html). You can also see [tabs in action in the Twitter sample](https://github.com/grahammendick/navigation/tree/master/NavigationReactNative/sample/twitter).

The default [Android transition animates the opacity which causes the white background to show through](https://github.com/grahammendick/navigation/pull/646). With a dark scene like yours you'll see a gray flash. Setting a transparent background on Android prevents the scene going gray.
```jsx
<NavigationStack
  backgroundColor={() => Platform.OS === 'android' ? 'rgba(255,255,255,0)' : 'white'}>
```

afkcodes:
thanks for the quick reply @grahammendick  i went through the earlier issues and tried this i think the opacity value was not there i will check this out. also i am unable to change the color of the unselected tab text color i have set 

unselectedTintColor ="white"

grahammendick:
Make sure you update to the latest `navigation-react-native` because it fixes the [unselectedTintColor not working on iOS](https://github.com/grahammendick/navigation/issues/682)

afkcodes:
i am on android 13.

grahammendick:
I set `unselectedTintColor` to 'red' in the Twitter sample on android and it's working. Can you try the same in [the Twitter sample](https://github.com/grahammendick/navigation/blob/5420aba1eca8097ddab298dda9d3f9ea0a7dcd2b/NavigationReactNative/sample/twitter/Tabs.js#L24), please?

![image](https://user-images.githubusercontent.com/1761227/229303591-73e2ce8c-0e68-40c9-8ead-7727d5ca1b9b.png)

afkcodes:
i am trying the same , but not able to run the twitter sample its giving me some error

grahammendick:
[...]

PR #699, opened 2023-04-09
Supported custom TabBar back behaviour (Android)

grahammendick:
Fixes https://github.com/grahammendick/navigation/issues/691#issuecomment-1493384062

@afkcodes wants the Android back button to step back through the visited tabs. Thought this was already possible by user adding a custom Android back handler. But the `TabBar` back handler took precedence because React Native added before the user's one (child components' effects run before parent's).

Added a `onPressBack` prop to the `TabBar` so the user can provide alternative back behaviour. 

For @afkcodes behaviour
```jsx
const [tab, setTab] = useState(0)
const tabsVisited = useRef([0]);
useEffect(() => {
  if (tab) {
    const i = tabsVisited.current.indexOf(tab);
    if (i > -1) tabsVisited.current.splice(i, 1);
    tabsVisited.current.push(tab);
  }
}, [tab]);
const handleBack = () => {
  const visited = tabsVisited.current;
  if (visited.length > 1) {
    visited.pop();
    setTab(visited[visited.length - 1]);
    return true;
  };
  return false;
}
return (
    <TabBar tab={tab} onChangeTab={setTab} onPressBack={handleBack}>
```

Or to close the app
```jsx
<TabBar onPressBack={() => false}>
```

### DIFF
--- a/NavigationReactNative/src/TabBar.tsx
+++ b/NavigationReactNative/src/TabBar.tsx
@@ -44,7 +44,7 @@ class TabBar extends React.Component<any, any> {
         return false;
     }
     render() {
-        var {children, labelVisibilityMode, barTintColor, selectedTintColor, unselectedTintColor, activeIndicatorColor, rippleColor, bottomTabs, scrollable, primary, scrollsToTop} = this.props;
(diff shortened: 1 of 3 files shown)


afkcodes:
This is so cool thanks @grahammendick  You are amazing

